PR #35, opened 2026-04-16
test: migrate test suite to stretchr/testify

ohalushchak-exadel:
## Summary
- Migrate all 35 test files across 10 packages to use `github.com/stretchr/testify` v1.11.1 (`assert` + `require`)
- Replace verbose `if err != nil { t.Fatal }` and `if got != want { t.Errorf }` patterns with concise testify assertions
- Use `require` for checks that must pass before subsequent code runs (nil guards, length checks before indexing), `assert` for independent checks
- Net reduction of ~1,100 lines of test boilerplate
- Add `internal/generate/generate` binary to `.gitignore`

Skipped benchmark-only and logging-only test files (no assertions to convert):
`bench/`, `e2e/perf_test.go`, `e2e/sizes_test.go`, `e2e/wire_bench_test.go`, `targeting/memory_test.go`, `targeting/scale_test.go`, `targeting/system_test.go`, `reference/context-agent/bench_test.go`

## Test plan
- [x] `go build ./...` and `go vet ./...` pass on all modules
- [x] All unit tests pass (2 pre-existing failures on `main` unrelated to this change: `registry/TestSyncer_AppliesAuthorizationRevoked`, `internal/generate/TestLoadSchemas`)


bokelley (approved):

bokelley:
much better

bokelley (changes requested):
## Review: testify migration

Good migration overall — the `require` vs `assert` split is correct in the critical paths (nil guards before dereferences, length checks before indexing). A few items to address before merging.

### Must fix

**1. `e2e/e2e_test.go` — two `json.Unmarshal` errors still silently dropped**

The PR wrapped most `json.Unmarshal` calls in `require.NoError`, but missed two (~lines 441, 495). Both are followed by `require.Len` on the unmarshaled struct, so a silent unmarshal failure produces a confusing "expected length 1, got 0" instead of the real error.

Fix: `require.NoError(t, json.Unmarshal(...))`

**2. `e2e/e2e_test.go` ~line 578 — type mismatch risk in `assert.Equal(t, uint64(1001), receivedRID)`**

testify uses `reflect.DeepEqual`, which is type-sensitive. If `receivedRID` is `int` or `int64`, this assertion will always fail even when the value is correct. Verify the type matches or cast consistently.

### Should fix

**3. Inconsistent `assert.Len` vs `assert.Equal(t, N, len(x))`**

~15 call sites use `assert.Equal(t, N, len(x))` where `assert.Len` would give better failure output ("expected length 3, got 5" vs "expected 3, got 5"). Since we're doing this migration once, worth standardizing on `assert.Len`.

**4. A few `t.Fatal` → `assert` downgrades where `require` would be more faithful**

Examples: `adcp/addtool_test.go` lines 93-94, `reference/context-agent/sync_test.go` line 45. None cause panics because there's no dependent code after them, but they weaken the stop-on-failure intent of the original.

bokelley (approved):

PR #39, opened 2026-04-16
refactor: extract ExposureRecorder from Engine for standalone library use

ohalushchak-exadel:
## Summary

Extract the frequency-capping and ad-exposure write path out of the `targeting` package into a new top-level `exposure` package. The targeting engine consumes `exposure` for read-side evaluation; systems that own impression recording can now embed `exposure.Recorder` against a bare Store without pulling in the engine, property registry, or signature machinery.

This replaces the earlier attempt (#?), which only split `RecordExposure` into a struct inside `targeting/`. Reviewers wanted a real package boundary so the dependency direction is visible — and so this seam becomes the natural place to later move exposure recording out of the identity-match library entirely (follow-up).

## Package layout

```
exposure/                   (new — no targeting import, no tmproto import)
├── exposure.go             ExposeRequest, ExposeResponse, UserIdentity, ExposureEntry/Log,
│                           UserProfile, FrequencyRule, validators, merge/prune/intent helpers
├── binary.go               BinaryExposureLog format + readers/writers
├── config.go               PackageIdentityConfig, CampaignFreqConfig, Load/BatchLoad/Seed
├── hash.go                 HashToken (SHA-256, user keys), HashString (FNV, compact IDs)
├── recorder.go             Recorder, RecorderConfig, Metrics, Clock, ClockFunc, RecordExposure
├── store.go                ConfigStore, BatchStore, RecorderStore interfaces
└── recorder_test.go        standalone-recorder test using a 30-line fakeStore — no targeting

targeting/                  (engine now imports exposure)
├── engine.go               uses exposure.HashToken, BatchLoadPackageIdentityConfigs,
│                           BinaryExposureLog, CheckFrequencyRulesMultiLog, etc.
├── context_config.go       PackageContextConfig loader (stays here — targeting-local)
├── metrics.go              ExposureRecorded removed; exposure.Metrics owns that callback
├── resolved.go             IdentityConfigs / CampaignConfigs now *exposure.* pointers
└── ...
```

**Dependency direction:** `targeting` → `exposure`. Never the reverse. `exposure` defines its own narrow Store interfaces; any `targeting.Store` satisfies them structurally, but `exposure` can also be wired against a minimal store with zero targeting coupling (demonstrated in `exposure/recorder_test.go`).

## What moved

| From `targeting` | To `exposure` |
|---|---|
| `ExposeRequest`, `ExposeResponse`, `UserIdentity`, `ValidateExposeRequest` | `exposure.*` (same names) |
| `ExposureEntry`, `ExposureLog`, JSON helpers, `Merge`/`Prune`/`Check`/`Latest`/`ComputeIntentScore` | `exposure.*` |
| `UserProfile`, `ParseUserProfile`, `MergeUserProfiles` | `exposure.*` |
| `FrequencyRule`, `FrequencyRuleJSON`, `ToFrequencyRules` | `exposure.*` |
| `PackageIdentityConfig`, `CampaignFreqConfig`, `Load*`, `BatchLoad*`, `Seed*` | `exposure.*` |
| `BinaryExposureLog` + all helpers | `exposure.*` |
| `HashToken` | `exposure.HashToken` |
| `hashString` (private) | `exposure.HashString` (now exported for cross-package access) |
| `Engine.RecordExposure` | `exposure.Recorder.RecordExposure` |

## Breaking changes

**`targeting.Metrics` no longer declares `ExposureRecorded`.** That callback moved to the new `exposure.Metrics` interface, which `exposure.Recorder` invokes.

External implementers who defined `ExposureRecorded` on their `targeting.Metrics` type will compile fine (Go interfaces ignore extra methods) but will silently stop receiving the callback once they upgrade. To continue receiving it, pass the same implementation as `exposure.RecorderConfig.Metrics`. See `docs/embedding.md` for the updated pattern — a single struct can satisfy both interfaces, as `targeting/prommetrics/Metrics` now does (compile-time asserted).

**Import-path renames for users of targeting-side types:**

| Old | New |
|---|---|
| `targeting.PackageIdentityConfig` | `exposure.PackageIdentityConfig` |
| `targeting.CampaignFreqConfig` | `exposure.CampaignFreqConfig` |
| `targeting.FrequencyRuleJSON` | `exposure.FrequencyRuleJSON` |
| `targeting.HashToken` | `exposure.HashToken` |
| `targeting.Seed{Package,Campaign}...Config` | `exposure.Seed{Package,Campaign}...Config` |
| `engine.RecordExposure(...)` | `exposure.NewRecorder(...).RecordExposure(...)` |
| `engine.Now = func() time.Time {...}` | `engine.Clock = exposure.ClockFunc(func() time.Time {...})` or pass `Clock` to `EngineConfig` |
| `mockStore.Now = ...` | `mockStore.Clock = ...` |

The reference identity-agent (`reference/identity-agent/cmd/...`), `targeting/valkeystore`, and `targeting/prommetrics` are all updated in this PR.

## Review-feedback addressed

From the previous reviews on the struct-only extraction:

- **Shadowing bug.** The `for _, r := range rules[1:]` loop inside the old `RecordExposure` shadowed the method receiver once the method moved onto a struct with receiver `r`. Renamed the loop variable to `rule`.
- **Shared clock.** `exposure.Clock` / `ClockFunc` replace the triple `Now func() time.Time` fields that previously lived on `Engine`, `MockStore`, and the `ExposureRecorder` draft. Tests now construct one `ClockFunc`, hand it to all three, and advance time by mutating the closed-over variable — no per-iteration reassignment. `TestSystem_RollingWindowExpiry` is the canonical example.
- **Standalone-recorder test.** `exposure/recorder_test.go::TestStandaloneRecorder_MinimalWiring` builds a `Recorder` from `(ProviderID, Store)` against a 30-line `fakeStore` that imports nothing from `targeting`. `TestStandaloneRecorder_SharedClock` verifies the shared-clock path.
- **Store-schema coupling called out.** Package doc on `exposure.Recorder` names the store keys (`user:exposures:<hash>`, `freq:pkg:*`, `freq:campaign:*`, `intent:*`) explicitly as the integration contract — any reader or writer honoring the schema interoperates.
- **Metrics breaking change documented.** `docs/embedding.md` now has an "Implementing targeting.Metrics and exposure.Metrics" section with the migration note.
- **Stale comment.** `tmproto/types_test.go:239` no longer claims ExposeRequest tests moved to `targeting`; they're in `exposure` now.

## Test plan

- [x] `go test ./...` (main module) — exposure / targeting / router / tmpclient / tmproto all green
- [x] `go test ./targeting/valkeystore/...` (miniredis integration) — green
- [x] `go test ./targeting/prommetrics/...` — green; both `targeting.Metrics` and `exposure.Metrics` conformance asserted
- [x] `go build ./reference/identity-agent/...` — green
- [x] `go test -run TestIntegration_ ./e2e/...` — green (integration round-trips context + identity + record over real router/registry/tmpclient)
- [x] `go vet ./...` — clean
- [x] Behavior of `RecordExposure` byte-identical to main (same store schema, same TTLs, same pruning math, same response shape — only change is the receiver rename and shadowing fix)

## Non-goals / follow-ups

- The reviewer flagged that if exposure recording is truly standalone, it probably shouldn't live in this repo at all — the natural owner is the impression pipeline (SSP / ad server / exchange callback), and the identity-match library should be read-only. This PR creates the package boundary that makes that relocation cheap; the actual move is a follow-up.

### DIFF
--- a/docs/embedding.md
+++ b/docs/embedding.md
@@ -51,38 +51,68 @@ r := router.NewRouter(providers, registry, nil, health,
 )
 ```
 
-## Implementing targeting.Metrics
+## Implementing targeting.Metrics and exposure.Metrics
 
-The `targeting.Metrics` interface is the observability hook for the evaluation engine. If your host application already has Prometheus (or any metrics system), implement the interface directly:
+Observability is split across two interfaces. The evaluation engine emits
+`targeting.Metrics` callbacks (context/identity evaluation outcomes, stage
+latency, store errors). The impression recorder in the `exposure` package
+emits `exposure.Metrics` callbacks (exposure recorded, store errors during
+write). A single type can satisfy both — that is what `targeting/prommetrics`
+does.
 
 ```go
 type MyMetrics struct {
-    contextEval *prometheus.CounterVec
-    latency     *prometheus.HistogramVec
-    // ...
+    contextEval     *prometheus.CounterVec
+    identityEval    *prometheus.CounterVec
+    exposureRecord  *prometheus.CounterVec
+    storeErrors     *prometheus.CounterVec
+    latency         *prometheus.HistogramVec
 }
 
+// targeting.Metrics — read path (evaluation).
 func (m *MyMetrics) ContextEvaluated(packageID, stage string, passed bool) {
     m.contextEval.WithLabelValues(stage, strconv.FormatBool(passed)).Inc()
 }
+func (m *MyMetrics) IdentityEvaluated(packageID, stage string, passed bool) { /* ... */ }
+func (m *MyMetrics) Latency(stage string, d time.Duration)                  { /* ... */ }
 
-func (m *MyMetrics) Latency(stage string, d time.Duration) {
-    m.latency.WithLabelValues(stage).Observe(d.Seconds())
+// Shared with exposure.Metrics — both interfaces include StoreError.
+func (m *MyMetrics) StoreError(operation string, err error) { /* ... */ }
+
+// exposure.Metrics — write path (impression recording).
+func (m *MyMetrics) ExposureRecorded(packageID string) {
+    m.exposureRecord.WithLabelValues().Inc()
 }
-
-// ... implement remaining methods
 ```
 
-Then pass it to the engine:
+Wire each interface where it's consumed:
 
 ```go
 engine := targeting.NewEngine(targeting.EngineConfig{
-    Metrics: &MyMetrics{...},
+    Metrics: &MyMetrics{...}, // satisfies targeting.Metrics
+    // ...
+})
+
+recorder := exposure.NewRecorder(exposure.RecorderConfig{
+    Metrics: &MyMetrics{...}, // satisfies exposure.Metrics
     // ...
 })
 ```
 
-The `targeting/prommetrics` sub-module is a reference implementation with zero external dependencies. Use it directly or as a template for your own.
+The `targeting/prommetrics` sub-module is a reference implementation with
+zero external dependencies that satisfies both interfaces on a single struct.
+Use it directly or as a template.
+
+### Breaking change in this split
+
+Prior versions exposed `ExposureRecorded` on `targeting.Metrics` because the
+engine owned impression recording. Recording is now owned by
+`exposure.Recorder`, and the callback lives on `exposure.Metrics`. External
+implementations that defined `ExposureRecorded` on their `targeting.Metrics`
+type still compile (extra methods don't break satisfaction), but the engine
+no longer invokes that method. To continue receiving the callback, pass the
+same implementation as `exposure.RecorderConfig.Metrics` when constructing
+the recorder.
 
 ## Security considerations for embedders
 
--- a/e2e/integration_test.go
+++ b/e2e/integration_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/adcontextprotocol/adcp-go/exposure"
 	"github.com/adcontextprotocol/adcp-go/router"
 	"github.com/adcontextprotocol/adcp-go/targeting"
 	"github.com/adcontextprotocol/adcp-go/tmpclient"
@@ -20,10 +21,11 @@ import (
 
 // testStack holds all components for an integration test.
 type testStack struct {
-	client         *tmpclient.Client
-	store          *targeting.MockStore
-	contextEngine  *targeting.Engine
-	identityEngine *targeting.Engine
+	client           *tmpclient.Client
+	store            *targeting.MockStore
+	contextEngine    *targeting.Engine
+	identityEngine   *targeting.Engine
+	exposureRecorder *exposure.Recorder
 }
 
 func setupStack(t *testing.T) *testStack {
@@ -41,8 +43,8 @@ func setupStack(t *testing.T) *testStack {
 	store.SetAdd("url:blocklist:pkg-family", targeting.HashURL("article:adult-content"))
 
 	// Seed audience segments.
-	aliceHash := targeting.HashToken("tok-alice")
-	bobHash := targeting.HashToken("tok-bob")
+	aliceHash := exposure.HashToken("tok-alice")
+	bobHash := exposure.HashToken("tok-bob")
 	store.SetAdd("audience:cooking_fans", aliceHash)
 	store.SetAdd("audience:sports_fans", bobHash)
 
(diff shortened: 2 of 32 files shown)


bokelley (commented):
Nice refactor overall — the lift is clean and tests were updated exhaustively. A few things worth discussing before merge.

## What's good
- Pure lift-and-shift: `RecordExposure` moves verbatim onto `ExposureRecorder`, receiver renamed `e`→`r`. No behavior drift.
- Narrowing `Metrics` by pulling `ExposureRecorded` into a focused 2-method `ExposureMetrics` is the right call — the write path doesn't need `ContextEvaluated`/`IdentityEvaluated`/`Latency`.
- Every `engine.RecordExposure` callsite migrated, including the valkey integration tests.

## Concerns

### 1. "Standalone library" oversells the decoupling
`RecordExposure` still calls `loadPackageIdentityConfig` and `loadCampaignFreqConfig`, and still leans on `resolveExposeIdentities`, `HashToken`, `BinaryExposureLog`, `maxExposureEntries`, `hashString`, `toFrequencyRules`. The PR description says "without configuring property registry, packages, signature verification" — true for the **constructor**, but the store must still be populated with package-identity and campaign-freq configs by something upstream, or the response's `CampaignCount`/`CampaignRemaining` will silently be zero. Worth calling that out in the description so importers aren't surprised.

### 2. Breaking change to `Metrics` not called out
Removing `ExposureRecorded` from the `Metrics` interface is a breaking API change. External `Metrics` implementers still compile (extra methods don't break satisfaction), but anyone who relied on the engine invoking `ExposureRecorded` on their `Metrics` impl will now get silence unless they *also* pass the same object as `ExposureMetrics` to `NewExposureRecorder`. A note in the PR body + CHANGELOG would save someone a confusing debug session.

### 3. Dual `Now` field is a papercut
Both `Engine` and `ExposureRecorder` expose public `Now` fields. Tests now have to set both at every time step — see `TestSystem_RollingWindowExpiry` where `engine.Now = …; recorder.Now = …; store.Now = …` appear in triplicate on every loop iteration. Easy to forget one and get silently wrong test time.

Consider (follow-up PR is fine):
- A shared `Clock interface{ Now() time.Time }` passed into both configs, or
- A `NewExposureRecorderFromEngine(e *Engine)` convenience that wires store + providerID + Now from the engine.

### 4. No test for the motivating use case
Every test builds both `Engine` **and** `ExposureRecorder` against the same store. None demonstrates "standalone recorder with zero engine configured" — which is literally the pitch of this PR. A ~15-line test (bare `MockStore` → `NewExposureRecorder` → `RecordExposure` → assert binary log entry + freq sorted-set member + source-hash fallback) would make the surface contract visible and prevent silent re-coupling later. Happy to sketch one if useful.

### 5. Variable shadowing carried over
In `RecordExposure`:
```go
shortestRule := rules[0]
for _, r := range rules[1:] {     // r shadows the method receiver
    if r.Window < shortestRule.Window {
        shortestRule = r
    }
}
```
Was benign before (receiver was `e`); now the loop var `r` shadows the `*ExposureRecorder` receiver. Rename to `rule`.

### 6. `testStack` alignment
In `e2e/integration_test.go`, the added `exposureRecorder *targeting.ExposureRecorder` field has a different column width than its neighbors — double-check `gofmt`'s alignment doesn't re-churn on CI.

## Verdict
Non-blocking — happy to see this land after:
- (required) fix the `r` shadowing,
- (required) add the standalone-recorder test,
- (nice-to-have) tighten PR description re: remaining store-schema coupling + `Metrics` breaking change.

Clock/shared-`Now` can be a follow-up.

bokelley:
Stepping back from the line-level review — I want to raise a bigger question before we invest more in this seam.

## Should `RecordExposure` live in this library at all?

The identity match server's job is to **read** frequency/exposure state to answer cap-check + audience-match questions. The writes happen upstream — whoever serves the impression (SSP, ad server, exchange callback) already has `impression_id`, `source_id`, `user_token`, and timing at the moment the ad is delivered. They're the natural owner of the write.

A few reasons this matters more than "cleaner package boundaries":

1. **TEE scope.** If the identity agent runs in a TEE, pulling `RecordExposure` into the enclave puts it on the ad-serving hot path with write authority to a shared store. That's scope creep against the whole point of the TEE (minimal, auditable surface). The enclave should be a pure function of `(request, store-state) → match result`.

2. **The real contract is the store schema, not the Go API.** `user:exposures:<hash>`, `freq:pkg:*`, `freq:campaign:*`, `intent:*` — that's what couples writers and readers. Any system honoring that schema is a valid writer. A Go function signature is the wrong abstraction boundary; it forces writers to embed our code, pick up our deps, and run our runtime.

3. **Separation of duties.** Reader = match engine. Writer = impression pipeline. Today they're fused into one library, and this PR is the first visible crack. The PR description implies "make it usable standalone" — but if it's going to be used standalone, it probably shouldn't be *here*.

## What I'd suggest instead

Reframe this PR (or a follow-up) as **step 1 of removing `RecordExposure` from the identity match library**:

- Move `exposure.go` (recorder + `ExposureMetrics` + `ExposureRecorderConfig`) into a separate package or sibling module — something like `exposure/` or a standalone SDK — so the import boundary is visible and the identity match library compiles without it.
- Leave `targeting/` with read-side only: `EvaluateIdentityResolved`, frequency cap checks, intent score. That's the TEE-safe surface.
- Document the store schema as the writer/reader contract.
- Open a follow-up to fully deprecate the recorder from this repo once the upstream exposure pipeline owns the writes end-to-end.

If we agree on that direction, the line-level nits in my earlier review (shadowing, standalone test, dual `Now`) mostly become moot — they'd be solved by the relocation, or they belong in whatever new home this code finds.

Curious what @ohalushchak-exadel and the team think. Happy to be wrong here if there's a reason the match server needs to own the write path I'm not seeing.

ohalushchak-exadel:
closing as it doesn't help us achieve what we were aiming for

PR #79, opened 2026-04-23
chore: Add DeleteUserProfile(s) method

ohalushchak-exadel:
Adds DeleteUserProfile and DeleteUserProfiles to Engine as counterparts to the existing setters, enabling callers to remove user segment profiles from the store.

  To support this, Del and MDel are added to the Store interface with corresponding implementations in MockStore and valkeystore.Store. The Valkey/Redis implementation uses a single DEL call for both the single and batch cases, since the command natively accepts multiple keys.


bokelley:
Triage status: still a potentially useful Store/Engine API addition, but not merge-ready. The branch conflicts with current main after substantial targeting/store evolution. Please rebase, update the Store implementations and tests to the current interfaces, and request a fresh review.

PR #100, opened 2026-04-26
feat(schemas): manifest-driven sync of canonical agent skills

### LINKED ISSUE #91: Sync canonical agent skills from the protocol tarball
## Context

[adcontextprotocol/adcp#3097](https://github.com/adcontextprotocol/adcp/pull/3097) (merged) bundles agent-facing skills (`skills/call-adcp-agent/SKILL.md` plus per-protocol task skills) into the published protocol tarball at `/protocol/<version>.tgz` alongside schemas and compliance. Companion PR [adcontextprotocol/adcp-client#965](https://github.com/adcontextprotocol/adcp-client/pull/965) wires the JS SDK to extract them on sync.

This issue tracks the equivalent for the Go SDK.

## What's in the tarball

The tarball's `manifest.json` now enumerates protocol-managed skills:

```json
{
  "contents": {
    "schemas": true,
    "compliance": true,
    "skills": ["adcp-brand", "adcp-creative", "adcp-governance", "adcp-media-buy", "adcp-si", "adcp-signals", "call-adcp-agent"],
    ...
  }
}
```

The seven directories live at `adcp-{version}/skills/<name>/`, each containing a `SKILL.md` (the agent-facing wire contract). Per-protocol skills also bundle a `schemas/` subdir which **should be filtered out** — the SDK already has those in its schema cache.

## Proposal

Extend the Go SDK's tarball-extraction path to:

1. After extracting the tarball, read `manifest.json`.
2. For each name in `manifest.contents.skills` (when present and an array), copy `<extract>/skills/<name>/` into the SDK's local `skills/<name>/`, **excluding nested `schemas/` subdirs**.
3. Snapshot the previous tree as `<dst>.previous` (matches the pattern already used for schemas/compliance).
4. Manifest-driven means SDK-local skills (anything not enumerated) stay untouched.

## Reference implementation

JS sync: [scripts/sync-schemas.ts in adcontextprotocol/adcp-client#965](https://github.com/adcontextprotocol/adcp-client/pull/965/files). The function `syncSkillsFromBundle` is ~30 lines; `copySkillTree` + `copyTreeFiltered` add another ~30. Translate to Go idiomatically (likely `filepath.Walk` with a skip-dir predicate).

## Why now

Without this sync, the buyer-side `call-adcp-agent` skill — the canonical wire contract for idempotency replay, account oneOf variants, async `submitted` polling, and error recovery via `adcp_error.issues[]` — doesn't reach Go SDK consumers. Coding agents using Go-flavored AdCP buyers fall back to training-data hallucinations of the wire shape, with predictably bad consequences for any mutating call (the oneOf account variant in particular has a high false-positive rate when LLMs guess at it).

Note: Go is the slowest of the three SDKs to evolve here, but it's also the one most likely to be embedded into automation pipelines where silent skill drift is hardest to debug. Worth catching up.

## Out of scope

- Adding new skills (separate issues upstream)
- Changing the manifest format (already settled upstream)

Co-tracked: adcontextprotocol/adcp-client-python (same ask, different SDK).

bokelley:
Closes #91.

## Summary

- `adcp/schemas/download.sh` now reads `manifest.contents.skills` after extracting the bundle and rsyncs each named skill directory into `/skills/<name>/`. Per-skill `schemas/` subdirs are excluded — verified byte-identical to the top-level schemas already in the SDK's schema cache.
- Skill names are validated against `^[a-zA-Z0-9_-]+$` before any `mkdir`/`rsync`. The validator emits zero output on failure (no partial list leak) and the script's `set -e` propagates the non-zero exit.
- Bundles without a `skills` entry no-op gracefully — SDK-local skills (`build-*`) are untouched in either path.
- `.github/CODEOWNERS` gates `/skills/` and `adcp/schemas/download.sh` behind owner review. Sigstore attests the bundle's origin; CODEOWNERS gates the file content (LLM wire contracts).

## Why no skill files in this PR

The current pin (3.0.0 GA) does not bundle skills — `adcp#3097` added bundling but that change is only in the `latest` dev snapshot. Pulling unsigned `latest` content under a "3.0.0 verified" pin would mix trust states. Skills will sync on the next `VERSION` bump to a release that bundles them.

## Triage blockers (from #91 comments)

1. **Per-skill `schemas/` filter correctness** — verified against `latest.tgz`: every per-skill schema is byte-identical to the corresponding top-level schema. Filter is correct.
2. **Skill name validation** — `^[a-zA-Z0-9_-]+$` enforced fail-closed.
3. **No `.previous` sidecar** — rsync `--delete` only, consistent with the existing schema sync pattern.
4. **CODEOWNERS** — `/skills/` and `download.sh` now require owner review.

`check-freshness.sh` is unchanged: 3.0.0 has no skills bundled, so there's nothing to drift-check yet. The bundle SHA already covers integrity at sync time. A skill-level drift linter (paralleling `lint.py` for schemas) follows when content lands.

## Test plan

- [x] `./download.sh latest` — syncs all 7 skills, filters per-skill `schemas/`, leaves `build-*` untouched
- [x] `./download.sh 3.0.0` — no-ops on skills (graceful skip), schemas sync as before
- [x] Tampered manifest with `../etc/passwd` — fail-closed, no partial sync
- [ ] Re-sync once 3.1.0 ships with skills bundled

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -25,3 +25,8 @@
 # Release tooling.
 /release-please-config.json           @bokelley
 /.release-please-manifest.json        @bokelley
+
+# Agent skills — wire contracts consumed verbatim by LLM agents. Sigstore
+# verifies the bundle's origin; CODEOWNERS gates the file content.
+/skills/                    @bokelley
+/adcp/schemas/download.sh   @bokelley
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,7 +47,7 @@ The targeting engine (`targeting/`) is the shared evaluation core. Reference age
 | `adcp/schemas/generate.py` | JSON-schema-to-Go-struct generator. `KNOWN_TYPES` set (top of file) lists types it skips so they can be hand-written. |
 | `adcp/schemas/lint.py` | Drift linter — diffs every `KNOWN_TYPES` entry against its schema. CI fails on drift. |
 | `adcp/testcontroller.go` | `RegisterTestController` — comply_test_controller for storyboard testing. |
-| `skills/` | SKILL.md files for coding agents to generate AdCP agents (seller, signals, creative, generative-seller, retail-media, collection). |
+| `skills/` | SKILL.md files for coding agents. SDK-local (`build-*`) authored here; protocol-managed (`adcp-*`, `call-adcp-agent`) synced by `adcp/schemas/download.sh` from the published bundle — do not hand-edit. CODEOWNERS-gated. See `skills/README.md`. |
 
 ### Adding a new AdCP type
 
--- a/adcp/schemas/download.sh
+++ b/adcp/schemas/download.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 # Downloads AdCP JSON schemas from the published protocol bundle at the pinned version.
+# Also syncs protocol-managed agent skills (LLM wire contracts) listed in
+# `manifest.contents.skills` into the repo's top-level `skills/` tree.
 #
 # Usage:
 #   ./download.sh              # download at pinned VERSION
@@ -198,7 +200,9 @@ PY
       # Exclude the per-skill schemas/ subdir at the skill root. These are
       # byte-identical copies of the top-level schemas already in the SDK's
       # schema cache; including them would duplicate ~1.4MB per protocol.
-      rsync -a --delete --exclude='/schemas/' "$SRC_SKILL/" "$DST_SKILL/"
+      # Anchored to skill root: drop a top-level `schemas` entry whether dir,
+      # file, or anything else. Nested foo/schemas/ trees are preserved.
+      rsync -a --delete --exclude='/schemas' "$SRC_SKILL/" "$DST_SKILL/"
       echo "Synced protocol skill: $SKILL_NAME"
     done < "$SKILL_LIST"
   fi
--- a/adcp/schemas/download.sh
+++ b/adcp/schemas/download.sh
@@ -153,6 +153,57 @@ rsync -a --delete "${EXCLUDE_ARGS[@]}" "$SRC/" "$SCRIPT_DIR/"
 # the upstream artifact (meaningful for the "latest" dev snapshot).
 echo "$EXPECTED" > "$SCRIPT_DIR/.bundle-sha256"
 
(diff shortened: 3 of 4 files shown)


bokelley:
## Expert review fixups (38d7fc0)

Ran the diff through `security-reviewer`, `code-reviewer`, `ad-tech-protocol-expert`, and `dx-expert`. No blockers — applied three fixes; logged four follow-ups.

### Applied
- **rsync exclude tightened** (`/schemas/` → `/schemas`): the trailing slash made the pattern dir-only, so a non-dir entry named `schemas` at a skill root would have passed through. Symlinks are already pre-rejected at line 133, so blast radius was small, but defense-in-depth costs nothing.
- **download.sh header now mentions skills sync**: contributors refreshing schemas no longer get surprised by `git status` churn under `/skills/`.
- **`skills/README.md` + AGENTS.md row**: documents the SDK-local (`build-*`) vs protocol-managed (`adcp-*`, `call-adcp-agent`) split. This is also the convention-based mitigation for the name-collision risk — manifest is human-curated upstream, current names don't overlap, runtime sentinels deferred.

### False alarm
- `code-reviewer` flagged `print("\n".join(names))` as dropping the last skill. Verified empirically — `print` adds the default `end='\n'` so the output is `a\nb\nc\n` and the bash `while read` loop reads all entries. No fix needed.

### Follow-ups (not blocking; will land alongside the first protocol-managed skill content)
1. **Sync-time per-skill schemas verification**: today the sync trusts the upstream invariant that per-skill `schemas/` are byte-identical to top-level. A future skill could legitimately ship an overlay and the SDK would silently drop it. Cheap fix: diff `$SRC_SKILL/schemas` against `$SCRIPT_DIR` before excluding, fail closed on divergence.
2. **Skill-drift detection in `lint.py` / `check-freshness.sh`**: catch local hand-edits to protocol-managed skill files before the next sync silently reverts them.
3. **`call-adcp-agent` SDK conformance audit**: separate work — diff SKILL.md replay/oneOf/submitted-polling semantics against `adcp/idempotency/` and `adcp/types.go` flatteners.
4. **Cross-SDK consistency check**: diff this Go impl against `adcp-client#965` (JS) — destination layout, schemas-overlay handling, manifest-absent fallback. Reconcile before consumers using both SDKs see divergent `skills/` trees.

PR #103, opened 2026-04-28
perf(targeting): preaggregate exposure log per filter hash for high-package eligibility

bokelley:
## Summary

Eligibility evaluation in `EvaluateIdentityResolved` re-scanned the user's exposure log per candidate package via `CheckFrequencyRulesMultiLog` and `LatestExposureMultiLog`, giving `O(packages × log_entries × identities)` CPU per request. At realistic Scope3-shape load (1000 candidate packages × 1000-entry log × 3 identities) this was ~7.5ms CPU per request; the pathological tail (1000 × 10K × 3) hit ~58ms — outside the 30ms p95 latency budget called out in the TMP spec.

This PR pre-buckets the user's exposure log entries by filter hash (campaign and package) **once per request**, plus precomputes per-package latest timestamp for intent score. Per-package eligibility check then walks only the matching bucket instead of the full log. Build cost `O(L × I)`, per-package check `O(matches-per-filter)`, independent of candidate-package count.

## Heuristic-gated, not always-on

`ShouldPreaggregate(numPackages) > 50` decides which path runs. Below that threshold, the map-build allocation overhead dominates — at 10 packages × 10K-entry log × 3 identities, the build cost more than triples per-request CPU (~1.25ms vs ~408µs naive). Above ~50 packages, preagg amortizes — at 1000 × 1000 × 3, ~26× speedup; at 1000 × 10K × 3, ~40× speedup.

The phase transition is sharp enough that a simple package-count check beats more elaborate heuristics. The two-path engine code is justified by avoiding a real measured regression on small-package requests.

## Measured speedups

vs. main, from `TestScale_IdentityMatch_CPU_Combined`, in-memory mock store, single goroutine:

| packages | log entries | identities | before    | after    | path   | speedup |
|----------|------------:|-----------:|----------:|---------:|--------|--------:|
| 10       | 10000       | 3          | 1,299 µs  | 898 µs   | naive  | 1.4×    |
| 1000     | 100         | 3          | 784 µs    | 71 µs    | preagg | 11.0×   |
| 1000     | 1000        | 3          | 7,566 µs  | 287 µs   | preagg | 26.4×   |
| 1000     | 10000       | 3          | 57,861 µs | 1,500 µs | preagg | ~38×    |

The pathological-tail case drops from 58ms to 1.5ms — comfortably within the latency budget.

## Bit-identical behavior

Same dedup (impression hash), same window filter, same MaxCount short-circuit. The naive `CheckFrequencyRulesMultiLog` and `LatestExposureMultiLog` functions remain in `exposure_binary.go` as public API with their existing tests; the engine just calls into the aggregated path when the threshold is exceeded.

## Test plan

- [x] `go test ./targeting/` passes
- [x] `go test ./targeting/ -run TestPreaggregate_Crossover` shows the threshold-driven crossover empirically
- [x] `go test ./targeting/ -run TestScale_IdentityMatch_CPU_Combined` confirms speedups
- [ ] Reviewer sanity-check on the threshold value (50) for any in-house workload that might want it tuned
- [ ] Production benchmarks against real valkey topology (out of scope; tracked in adcp#3359 rollout plan)

Related: [adcontextprotocol/adcp#3359](https://github.com/adcontextprotocol/adcp/pull/3359) — IdentityMatch architecture spec that surfaced this perf concern.

### DIFF
--- a/targeting/cpu_combined_test.go
+++ b/targeting/cpu_combined_test.go
@@ -1,0 +1,110 @@
+package targeting
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/adcontextprotocol/adcp-go/tmproto"
+)
+
+// TestScale_IdentityMatch_CPU_Combined measures EvaluateIdentityResolved CPU
+// across the combined dimensions that matter for production sizing:
+// (candidate packages per request) × (exposure log entries per identity) ×
+// (identities per request). All numbers are isolated from network I/O via
+// the mock store, so they represent in-process CPU only.
+func TestScale_IdentityMatch_CPU_Combined(t *testing.T) {
+	pkgCounts := []int{10, 100, 1000}
+	logSizes := []int{0, 100, 1000, 10000}
+	identityCounts := []int{1, 3}
+
+	t.Log("")
+	t.Log("=== IdentityMatch CPU: packages × log_size × identities ===")
+	t.Log("")
+	t.Logf("  %-10s %-10s %-10s %-15s %-12s", "packages", "log_size", "identities", "ns/op", "µs/eval")
+	t.Logf("  %-10s %-10s %-10s %-15s %-12s", "--------", "--------", "----------", "-----", "-------")
+
+	for _, numPkgs := range pkgCounts {
+		for _, logSize := range logSizes {
+			for _, numIdentities := range identityCounts {
+				now := time.Date(2026, 1, 1, 12, 0, 0, 0, time.UTC)
+				store := NewMockStore()
+				store.Now = func() time.Time { return now }
+
+				// Set up N packages, each with one fcap rule of max=10/86400s.
+				var pkgs []PackageConfig
+				var pkgIDs []string
+				idConfigs := make(map[string]*PackageIdentityConfig, numPkgs)
+				for i := range numPkgs {
+					pkgID := fmt.Sprintf("pkg-%d", i)
+					pkgs = append(pkgs, PackageConfig{PackageID: pkgID})
+					pkgIDs = append(pkgIDs, pkgID)
+					idCfg := PackageIdentityConfig{
+						FrequencyRules: []FrequencyRuleJSON{{MaxCount: 1_000_000, WindowSeconds: 86400}},
+					}
+					store.SetPackageIdentityConfig(pkgID, idCfg)
+					idConfigs[pkgID] = &idCfg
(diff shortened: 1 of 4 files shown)


targeting/exposure_aggregate_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
+// TestPreaggregate_Crossover measures naive vs preaggregated frequency-cap
+// evaluation across the (packages × log_entries × identities) matrix to
+// determine where the heuristic threshold should sit.
+func TestPreaggregate_Crossover(t *testing.T) {
```
ohalushchak-exadel, line 12: this doesn't really test anything, so should not be run together with the real tests. (should utilize `t.Skip()`)

targeting/exposure_aggregate_test.go
```diff
@@ -0,0 +1,84 @@
+package targeting
+
```
ohalushchak-exadel, line 2: this needs a test for `CheckFrequencyRulesMultiLog(...) == CheckFrequencyRulesAggregated(BuildPreaggregatedExposures(...), ...)` for identical inputs

targeting/exposure_aggregate_test.go
ohalushchak-exadel: or ideally this would be a `Benchmark`, not `Test`

targeting/cpu_combined_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/adcontextprotocol/adcp-go/tmproto"
+)
+
+// TestScale_IdentityMatch_CPU_Combined measures EvaluateIdentityResolved CPU
+// across the combined dimensions that matter for production sizing:
+// (candidate packages per request) × (exposure log entries per identity) ×
+// (identities per request). All numbers are isolated from network I/O via
+// the mock store, so they represent in-process CPU only.
+func TestScale_IdentityMatch_CPU_Combined(t *testing.T) {
```
ohalushchak-exadel, line 17: same as `TestPreaggregate_Crossover` - no real tests, should be skipped or `Benchmark`

bokelley:
Closing per the decision already recorded on #104: the measured BucketStore plus RollingStore direction supersedes this binary-log preaggregation path. This branch also conflicts with main and its last CI failed, so it is not a viable merge candidate.

PR #114, opened 2026-05-08
feat(tmp): Ed25519 request signing + HPKE TMPX exposure tokens

ohalushchak-exadel:
## Summary

Two TMP spec gaps closed end-to-end in the same branch, plus the v3.0.7 schema bump and the security/code-review hardening pass:

1. **Request authentication** ([spec §Request Authentication](https://adcontextprotocol.org/docs/trusted-match/specification#request-authentication)) — Ed25519 signing on every router→provider fan-out (`X-AdCP-Signature` / `X-AdCP-Key-Id`), per-provider binding, daily-epoch replay window. The previous `tmproto/signing.go` predated the current spec (wrong field set, no `provider_endpoint_url` binding, no JCS) and the router/reference-agents sent and accepted unsigned requests. The misleading `"TMP no longer defines request-level signing"` comment in `router/router.go` is removed.
2. **TMPX exposure tokens** ([spec §TMPX Exposure Tokens](https://adcontextprotocol.org/docs/trusted-match/specification#tmpx-exposure-tokens)) — HPKE `mode_base` with the spec-fixed cipher suite (DHKEM(X25519, HKDF-SHA256) / HKDF-SHA256 / ChaCha20-Poly1305). Until now `IdentityMatchResponse.Tmpx` was defined in the schema but no agent in the repo populated it.
3. **AdCP schema bundle bumped to v3.0.7** (Sigstore-verified) — `types_gen.go` is the version-header bump; the wire-level change is in `call-adcp-agent` skill error-envelope semantics (`property_name` instead of `field` inside `discriminator`, `hint` / `allowedValues` reclassified as SDK-side enrichment).

### Protocol layer (`tmproto/`)
- **`signing.go`** — newline-joined signing input for context match; `hex(SHA-256(JCS(canonical_object)))` for identity match; per-provider URL binding; daily-epoch replay window (current + previous); revocation via `revoked_at`. `Signer.PrivateKey` is unexported — derived material flows only through `PublicJWK()`.
- **`jcs.go`** — stdlib-only RFC 8785 (JSON Canonicalization Scheme). Object keys sort by UTF-16 code units. Non-integer floats are rejected (Go's `strconv.FormatFloat 'g'` is not exact ECMA-262; loud failure prevents future drift).
- **`verify_middleware.go`** — `VerifyContextMatchHandler` / `VerifyIdentityMatchHandler` for reference providers. Decoder runs with `DisallowUnknownFields` so a future-protocol field can't silently disappear from the recomputed signing input.
- **`keystore_remote.go`** — polls the router's `/registry/snapshot` for signing-key discovery, 5-min TTL. **Hardened:** `https://` required by default (opt-in `AllowInsecureScheme` for dev), cross-origin redirects denied, body capped at 1 MB, transient empty snapshots retain cached keys, cross-property `kid` collisions warn-and-keep-first. Lifecycle is `Refresh` + `Run` (no double-Start).
- **`tmpx.go`** — HPKE mode_base on stdlib (`crypto/ecdh`, `crypto/hkdf`, `crypto/sha256`) + `golang.org/x/crypto/chacha20poly1305`. No third-party HPKE framework. Binary-plaintext encoder (16-byte header + entries), TMPX type-ID registry, `kid.base64url(enc||ct)` wire format. `labeledExpand` validates length ≤ 0xffff.

### Router (`router/`)
- `WithTMPSigner` option. Signature headers attached on every fan-out, identity-match re-signed per provider.
- New `signing.go` — `(placement_id, provider_endpoint_url, epoch)` cache for context-match signatures.
- `RegistryProperty.SigningKeys`, `AttachSigningKey`, kid-indexed `LookupKey` with first-seen-wins on cross-property `kid` collisions.
- `HandleContextMatch` calls `Artifact.StripAccess()` before fan-out (spec MUST — was a TODO before this PR).
- `HandleIdentityMatch` surfaces `json.Marshal` failure as 500 with `request_id` instead of fanning out a stale body.
- Fixed misleading comment at lines 73-74.

### Wiring
- `cmd/router/main.go` — fail-closed unless `TMP_ROUTER_SIGNING_DISABLED=true`. Attaches public JWK to authorized property RIDs in the registry so providers can fetch it via `/registry/snapshot`.
- **Reference identity-agent**: `--registry-url`, `--allow-unsigned` (default off — verification is required), `--own-endpoint-url`. TMPX flags `--tmpx-kid`, `--tmpx-pubkey-path`, `--tmpx-country` plus required ack `TMP_IDENTITY_TMPX_REFERENCE_STUB_ACK=1` (the SHA-512 binary-token stub is non-interoperable with real buyer masters and operators must acknowledge before the agent will start).
- **Reference context-agent**: same flag set, same default-on signature verification.
- Both agents honor flag > env > default precedence (was env-can-set-but-not-clear before).
- `RemoteKeyStore.Run` is driven by a process-lifetime context; previous double-`Start` race fixed.

### Docs
- `docs/network-surface.md`:
  - Ed25519 section: `X-AdCP-*` header names, signed-input shapes, JCS for identity, per-provider binding, **revocation grace** (~24h via the previous-epoch window made explicit).
  - TMPX section: cipher suite, wire format, plaintext layout, env vars, the reference-impl stub callout.
  - Crypto agility note: one suite at a time; upgrade requires editing constants and dispatching by `kid` prefix or JWK fields.
  - Keystore hardening (HTTPS-default, redirect denial, body cap, kid-collision policy).

## Decisions

- **Replaced `tmproto/signing.go` API** rather than versioning. Only callers were the context-agent benchmarks; updated in this PR.
- **Hand-rolled HPKE** rather than pull a framework. Spec fixes one cipher suite; the implementation is ~250 lines, stdlib-only beyond `chacha20poly1305`. Validated against the **RFC 9180 §A.3 test vector** byte-exact at every KDF stage (`enc`, `shared_secret`, `secret`, `key`, `base_nonce`, ciphertext).
- **Registry-based key distribution** per spec direction. Router publishes its public JWK on the property records it's authorized to sign for; agents fetch via the existing `/registry/snapshot` endpoint with a 5-minute TTL.
- **Reference agents default to fail-closed** on signature verification. `--allow-unsigned` is the explicit opt-out for migration windows. The TEE-bound reference impl shouldn't normalize "no auth" as the path of least resistance.
- **TMPX SHA-512 stub gated behind `TMP_IDENTITY_TMPX_REFERENCE_STUB_ACK=1`.** Real buyer deployments decode UID2 / RampID / MAID per the source graph's encoding; the stub produces tokens no buyer master can decode. Acknowledgment env var prevents accidental production wiring with zero buyer match-rate.
- **No sample-rate verification** — verifier runs at 100% in v1. Spec says SHOULD; revisit if perf demands.

## Test plan

- [x] `go test ./...` clean across `tmproto`, `router`, `cmd/router`, `reference/context-agent`, `reference/identity-agent`, `e2e`, `adcp`, `bench`, `targeting/*`.
- [x] **RFC 9180 §A.3 vector** validated byte-exact at every KDF stage.
- [x] HPKE Seal/Open roundtrip across many random plaintexts.
- [x] Signing roundtrip: signer → middleware verifier across context match and identity match.
- [x] Wrong `provider_endpoint_url` rejected (per-provider binding).
- [x] Stale epoch rejected (>1 day).
- [x] Revoked key rejected.
- [x] Identity dedup + sort produce deterministic signing inputs regardless of input order.
- [x] Identity-match signatures differ across providers (per-provider re-signing).
- [x] Context-match signature cache reuses across requests in the same epoch.
- [x] TMPX plaintext header layout (version, ts, country, nonce, count) verified at the byte level.
- [x] TMPX type-ID registry sizes (32 / 48 / 16 bytes) match spec.
- [x] Unknown TMPX type IDs rejected at encode time; unmappable `uid_type` values dropped at the agent (forward-compat).
- [x] **`Artifact.StripAccess` runs before context-match fan-out** — bearer tokens never reach providers (test asserts on the forwarded bytes).
- [x] **`RemoteKeyStore` rejects non-HTTPS URLs by default**, denies cross-origin redirects, retains cached keys on transient empty snapshots.
- [x] **Cross-property `kid` collisions** warn and keep first-seen entry in both the registry index and the remote keystore parser.
- [x] **`Run` returns `context.Canceled`** when its context is canceled (no goroutine leak).
- [x] **JCS rejects non-integer floats** (test asserts on `1.5`); accepts integer floats.
- [x] Lint clean across `tmproto`, `router`, `cmd/router`, `reference/{identity,context}-agent`, `e2e`.


router/signing.go
```diff
@@ -0,0 +1,120 @@
...
+type contextSignatureCache struct {
+	mu         sync.Mutex
+	entries    map[contextSignatureCacheKey]string
+	maxEntries int
+}
+
+type contextSignatureCacheKey struct {
+	placementID string
+	endpointURL string
+	epoch       int64
```
BaiyuScope3, line 34: Very smart 👍 ! The Ed25519 signing is fast but not free, using cache is a good call.

However don't forget Ed25519 signing is deterministic and bound to the message,  in another word, if message_bytes is different,  the client side verification will be fail to varify. 

so if we only have `placementID`, `endpointURL`, `epoch` as cache key, it means 2 requests as long as they're having the same 3 columns they will use the same signing **without** checking `package_ids` from the cache

req1 sign with package_id(1,2) cannot directly use to req2 's package_id(3) since the message will be different, the client will fail on the signature verification. :)

ohalushchak-exadel: Good catch — fixed in 5cd71b3.

You're right that the cache key needs to cover `package_ids` to be sound. The spec MUST on `ContextMatchRequest.PackageIDs` ("MUST NOT vary by user — the same package_ids must be sent for every user on a given placement") makes the original key correct for spec-compliant traffic, but since the publisher controls `package_ids`, leaving it out of the key turns a spec violation into a signature/body mismatch that providers reject with no obvious upstream cause. Including it converts that case into a transparent cache miss instead.

The fix adds `packageIDsKey()` that sorts and comma-joins the same way the signing input does, so distinct sets get distinct entries and two requests differing only in input order still share one entry. New tests assert both halves.

docs/network-surface.md
```diff
@@ -94,6 +95,33 @@ Exposure tracking uses encrypted TMPX tokens instead of a dedicated endpoint:
...
+| Country | 2 | ISO 3166-1 alpha-2, ASCII; data-residency hint, buyer-internal |
+| Nonce | 8 | Random; deduplication at the master |
+| Count | 1 | Number of identity entries |
+| Entries | variable | `type_id (1 byte) + token (size from registry)` |
+
+**Reference identity-agent configuration:**
+
+| Flag / env var | Purpose |
+|---|---|
+| `--tmpx-kid` / `TMP_IDENTITY_TMPX_KID` | Buyer-cluster recipient kid (≤8 chars) |
```
BaiyuScope3: Let' remove the `tmpx-kid` from the config, and add `tmpx-enc-pubkey-ttl` instead
because the enc key will be rotated at anytime, the match service will need to load the `jwks.json` with filtering on adcp-use column = "tmpx-encrypt", and attach the kid in the tmpx
- example jwks file: https://api.staging.interchange.io/.well-known/jwks.json

docs/network-surface.md
```diff
@@ -141,24 +169,62 @@ The router tracks per-provider health:
...
+| `TMP_CONTEXT_REGISTRY` | Context Agent | Path to local registry snapshot | (none) |
+| `TMP_CONTEXT_REGISTRY_URL` | Context Agent | URL of router's `/registry/snapshot` for signing keys | (none) |
+| `TMP_CONTEXT_ENDPOINT_URL` | Context Agent | Own registered endpoint URL (signed-binding check) | (none) |
+| `TMP_CONTEXT_REQUIRE_SIGNATURE` | Context Agent | Reject unsigned requests | `false` |
 | `TMP_IDENTITY_ADDR` | Identity Agent | Listen address | `:8082` |
 | `TMP_IDENTITY_REDIS_ADDR` | Identity Agent | Valkey/Redis address | (none, uses in-memory) |
+| `TMP_IDENTITY_REGISTRY_URL` | Identity Agent | URL of router's `/registry/snapshot` for signing keys | (none) |
+| `TMP_IDENTITY_ENDPOINT_URL` | Identity Agent | Own registered endpoint URL (signed-binding check) | (none) |
+| `TMP_IDENTITY_REQUIRE_SIGNATURE` | Identity Agent | Reject unsigned requests | `false` |
+| `TMP_IDENTITY_TMPX_KID` | Identity Agent | Buyer-cluster TMPX recipient kid (≤8 chars) | (none) |
```
BaiyuScope3: same here let's dynamically load kid for encryption from the path
I would recomand to also add `encrypt` terms in the param name, so it will be much intuitive to know
- which one for signing
- which one for encryption

reference/identity-agent/cmd/identity-agent/main.go
```diff
@@ -222,3 +298,166 @@ func seedConfigs(store targeting.Store) (*targeting.ResolvedPackages, error) {
 		IdentityConfigs: idConfigs,
 	}, nil
 }
+
+// resolveString picks the configured value for a string flag with the
+// precedence flag > env > default. flagSet says whether the flag was passed
+// on the command line; only when it wasn't may an env var override.
+func resolveString(flagVal string, flagSet bool, envName string) string {
```
BaiyuScope3, line 306: should we have a load and verification function on the jwks.json :)
```
for _, k := range jwks.Keys {
    if k.Kty != "OKP" { skip }
    switch k.AdcpUse {
    case "request-signing":
        if k.Crv != "Ed25519" || k.Alg != "EdDSA" || k.Use != "sig" {
            skip + log
        }
        // add to signing ttl
    case "tmpx-encrypt":
        if k.Crv != "X25519" || k.Alg != "HPKE-DHKEM-X25519-HKDF-SHA256" || k.Use != "enc" {
            skip + log
        }
        // base64url-decode x → 32 bytes → ecdh.X25519().NewPublicKey
        // add to TMPX encryption ttl
    default:
        skip
    }
}
```

reference/identity-agent/cmd/identity-agent/main.go
```diff
@@ -222,3 +298,166 @@ func seedConfigs(store targeting.Store) (*targeting.ResolvedPackages, error) {
...
+// buildTmpxToken seals an HPKE TMPX token containing the resolved identities.
+// Identities whose UIDType has no TMPX type-ID mapping are dropped silently
+// per the spec's forward-compatibility rule (unknown types skipped).
+//
+// The string→binary conversion in stubBinaryToken is a reference stub —
+// real buyer deployments decode UID2/RampID/etc. according to the source
+// graph's encoding. Tokens produced here are not interoperable with a real
+// buyer master.
+func buildTmpxToken(cfg *tmpxConfig, ids []tmproto.IdentityToken) (string, error) {
+	entries := make([]tmproto.TmpxEntry, 0, len(ids))
```
BaiyuScope3: According to the spec: 
```
“When the buyer resolves more identities than fit the budget, the TMPX plaintext is truncated to the highest-priority entries according to buyer deployment configuration. The priority order is a buyer-side configuration concern (not protocol-level), typically ranking deterministic graphs (UID2, RampID) above probabilistic or publisher-scoped identifiers. Buyers MUST configure an explicit priority list — the default implementation MUST NOT truncate arbitrarily.”
```

We need to have: 
- Priority-list config flag
- Wire-size check before sealing
- Truncation logic
- Fail-loud-when-overflow path

right ?

just to think out loud
deployment config should have 
```
tmpx_priority_order:  # explicit ranking; highest first
  - uid2
  - rampid
  - rampid_derived
  - euid
  - id5
  - hashed_email
  - pairid
  - maid
  - publisher_first_party
max_wire_chars: 255
```

and in the encoder: 
1. we're sort by prioritity 
2.  Encode + seal a trial run. If wire > max, drop lowest-priority entry, retry

BaiyuScope3 (approved):

tmproto/keystore_jwks.go
```diff
@@ -0,0 +1,286 @@
...
+		case adcpUseRequestSigning:
+			if err := validateSigningJWK(&k); err != nil {
+				if logger != nil {
+					logger.Warn("jwks signing key skipped", "kid", k.Kid, "error", err)
+				}
+				continue
+			}
+			if _, dup := signing[k.Kid]; dup {
+				if logger != nil {
+					logger.Warn("jwks duplicate signing kid — keeping first-seen", "kid", k.Kid)
```
BaiyuScope3, line 228: we might not need public signing key here , but if we want to save the key(verification purpose), we should keep both signing kid in the mem, since we never know incoming request was signed with which one :)

ohalushchak-exadel: this specific case is a misconfiguration, we should not see the same kid twice in the same file

PR #117, opened 2026-05-12
feat(tmp): adopt seller_agent_url on identity match, regen schemas to latest

ohalushchak-exadel:
## Summary

Adopts the new identity-match wire contract by pinning `adcp/schemas` to `latest` and regenerating `tmproto`. Lays the groundwork for an in-memory targeting-config service (PR2) that needs `(seller_agent_url, package_id)` as the lookup key.

Key protocol changes regenerated into `tmproto/types_gen.go`:

- `IdentityMatchRequest.seller_agent_url` — new required field; buyer resolves the active package set per seller.
- `IdentityMatchRequest.package_ids` — now optional (evaluation defaults to the seller's full active set when omitted).
- `IdentityMatchResponse.ttl_sec` → `serve_window_sec` — semantic shift from router response-cache TTL to per-package single-shot fcap window.
- `adcp_version` (release-precision pin) added to both context and identity match requests.

## What changed

- **`tmproto/validate.go`** — require `seller_agent_url`; drop the empty-`package_ids` rejection.
- **`router/router.go`** — `mergeIdentityResponses` renames `minTTL` → `minServeWindow`. Mechanical min-across-providers preserved; whether the router should still use this value as a response cache key is a separate product decision, intentionally not changed here.
- **`tmpclient.ActivateParams`** — gains `SellerAgentURL` plumbed into the identity request build path.
- **All call sites** in `router/`, `tmpclient/`, `e2e/`, and `reference/identity-agent/` updated for the field rename and the new required field.
- **`TestIdentityMatchRequest_NoContextFields`** — substring guard against page-context leakage swaps `"url"` for `"page_url"` so the legitimate `seller_agent_url` field no longer trips it.

## What is NOT in this PR

- No changes to router response-cache key/TTL semantics under the new `serve_window_sec` meaning. Mechanical rename only.
- No in-memory targeting-config service. That's PR2 (waiting on this to merge).

## Test plan

- [x] `go test -count=1 ./...` in root, `adcp/`, `reference/identity-agent/`, `e2e/` — all pass in isolation.
- [x] Build clean across all three modules.
- [ ] Note: `TestSimulation_*` and `TestPerformance_EndToEnd/throughput` are pre-existing macOS port-exhaustion flakes when the full e2e suite runs in one process; they pass when run individually. Not introduced here.


adcp/schemas/VERSION
```diff
@@ -1 +1 @@
-3.0.7
+latest
```
BaiyuScope3: can we specify the adcp version
1. latest will include the proto changes from upperstream that are not version tagged (not released version)
2. it will also hard to tell which version the current adcp-go are implemented with
3. it might accidently deployed the proto change and break things unintentionally

BaiyuScope3: `3.0.11` was released yesterday

BaiyuScope3 (dismissed):
Don't forget to add new sales agent url in the BuildIdentityMatchSigningInput, to be signed

tmpclient/types.go
BaiyuScope3: nit: the formatting seems broken here

adcp/schemas/VERSION
ohalushchak-exadel: 3.0.11 didn't have the required changes, so moved to lates, but 3.0.12 was released today - switched to it

BaiyuScope3 (approved):

PR #118, opened 2026-05-13
feat(identityconfig): in-memory service + Scope3 HTTP source

ohalushchak-exadel:
## Summary

Introduces a new in-memory `targeting/identityconfig` package that serves PackageIdentityConfig entries keyed by `(seller_agent_url, package_id)` and an HTTP `scope3` implementation that posts to a configurable endpoint with Bearer auth. Wires both into the reference identity-agent and removes the prior Redis-backed identity-config loading path.

Follow-up to [#117](https://github.com/adcontextprotocol/adcp-go/pull/117) (the wire-contract upgrade that introduced `seller_agent_url`).

## What's in the package

- `Source` interface: `LoadAll` and `LoadUpdatedAfter(after)`. Snapshot carries `LastUpdatedAt`; Delta carries `Upserted`, `Removed`, and the new watermark.
- `Service` with `atomic.Pointer[snapshotData]` — readers (`Get`, `GetBySeller`) are lock-free; each refresh builds an immutable new snapshot and CAS-installs it. `Start` launches a background refresh tick; `Stop` drains it.
- `StartMode` selects initial-load behavior: `FailFast`, `Retry` (with `Initial`/`Max`/`Backoff`/`MaxAttempts`/`Deadline`), or `BestEffort` (proceed with empty snapshot, log error, retry on the normal tick).
- `ResolveRequest(svc, sellerAgentURL, packageIDs)` returns the effective evaluation set per the wire contract: empty `packageIDs` → seller's full registered set; otherwise intersection with unregistered IDs silently dropped to avoid leaking registry membership.
- `scope3.Source` — HTTP POST with Bearer auth, configurable URL + token + HTTP client/timeout. Snake-case JSON tags throughout; supports the agreed request `{after?}` body and the response shape with `last_updated_at` / `targeting_configs[]` / `removed_targeting_configs[]`.

## Schema reshape

`PackageIdentityConfig.TargetSegments` changes from `[]string` to `*SegmentRule{AllOf, AnyOf, NoneOf}` — a single AND-of-clauses rule. Empty/nil rule = no audience gating.

The engine's identity evaluation resolves the union of segments referenced across all requested packages, batch-queries audience membership once, then applies `SegmentRule.Matches(userSegments)` per package. Dropped `ResolvedPackages.SegmentIndex` and the related `SegmentCandidates` helper — the segment-shaped optimization doesn't fit the new shape and per-package matching is straightforward.

## Identity-agent wiring

New flags / env vars on the reference identity-agent:

- `--config-source-url` / `TMP_IDENTITY_CONFIG_SOURCE_URL` (required)
- `--config-source-token` / `TMP_IDENTITY_CONFIG_SOURCE_TOKEN` (required)
- `--config-source-timeout` (default 30s)
- `--config-refresh-interval` (default 5m)

The agent constructs the scope3 source, builds the service with `StartModeRetry` (1s initial → 30s max, exponential, 5min deadline), and starts it before serving. Per request, `identityconfig.ResolveRequest` resolves the effective package set; a tiny `ResolvedPackages{IdentityConfigs: ...}` is built per request and passed to `engine.EvaluateIdentityResolved` as before.

## Redis cleanup

Removed identity-config Redis paths: `batchLoadPackageIdentityConfigs`, `SeedPackageIdentityConfig`, `MockStore.SetPackageIdentityConfig`, and the dead `config:pkg:{packageID}` seed lines in glidestore/redisstore integration tests. Context configs and other Redis usage are untouched.

## Concurrency

Service uses `atomic.Pointer` swap — readers never block, never coordinate with refreshes. Each refresh allocates a fresh map and CAS-installs it. Refresh failures keep the prior snapshot. `Service.Stop` cancels the ticker context and joins the goroutine.

## Test plan

- [x] `go test -race ./...` — root, e2e, reference/identity-agent, adcp all green.
- [x] New unit tests cover: initial load, retry/exhaustion, BestEffort, delta upsert+remove, double-start rejection, idempotent Stop, ResolveRequest intersection + empty-IDs branches.
- [x] scope3 tests cover: Bearer header, request body shape (with and without `after`), response parsing with full rule, HTTP error surfacing, context cancellation.

## Out of scope

- The `SegmentIndex` field was unused outside identity eval. If a future context-side use emerges, it can be reintroduced.
- 3.0.x `ttl_sec` → 3.1.0 `serve_window_sec` and the new `adcp_version` field still wait on the next upstream cut.


targeting/identityconfig/service.go
```diff
@@ -0,0 +1,364 @@
...
+	s.mu.Unlock()
+
+	if err := s.initialLoad(ctx); err != nil {
+		s.mu.Lock()
+		s.running = false
+		s.mu.Unlock()
+		return err
+	}
+
+	loopCtx, cancel := context.WithCancel(context.Background())
```
BaiyuScope3: If there is a `Stop` during initial load, it won't actually cancels the retry. 
Consider wire the cancel before the load. Move the `loopCtx, cancel := context.WithCancel(...)` block up so s.cancel exists before initialLoad is invoked. Then make loadAllWithRetry’s select include <-loopCtx.Done()

targeting/engine.go
```diff
@@ -313,58 +313,31 @@ func (e *Engine) EvaluateContextResolved(ctx context.Context, resolved *Resolved
...
 	for _, pkgID := range req.PackageIDs {
 		idCfg := resolved.IdentityConfigs[pkgID]
 		eligible := true
 
-		if idCfg != nil && len(idCfg.TargetSegments) > 0 {
-			if _, ok := segmentEligible[pkgID]; !ok {
+		if idCfg != nil && !idCfg.TargetSegments.IsEmpty() {
+			if e.audience == nil || !idCfg.TargetSegments.Matches(userSegments) {
 				eligible = false
 				e.metrics.IdentityEvaluated(pkgID, StageAudience, false)
```
BaiyuScope3: I have cardinality concern using per-pkgID (seems unbounded) will makes very high cardinality , do we know If buyer-side pkgIDs are reasonably bounded? if not we probably shouldn't keep pkgID here

targeting/identityconfig/service.go
ohalushchak-exadel: makes sense, fixed

targeting/engine.go
ohalushchak-exadel: existing implementation ignored it already - dropped the param completely

targeting/identityconfig/service.go
```diff
@@ -0,0 +1,385 @@
...
+}
+
+// loadAllWithRetry retries until success, attempts/deadline exhausted, or
+// the context is cancelled.
+func (s *Service) loadAllWithRetry(ctx context.Context) error {
+	cfg := s.start.Retry
+	if cfg.Initial <= 0 {
+		cfg.Initial = time.Second
+	}
+	if cfg.Max <= 0 {
```
BaiyuScope3: nit: This (<=0)check is not necessary right? 
since l240 garenteed Initial >0
and l246 will covered <= 0 isn't it ?

ohalushchak-exadel: yup

targeting/identityconfig/service.go
```diff
@@ -0,0 +1,384 @@
...
+// Entry order is implementation-defined and may differ between snapshots;
+// callers needing a stable order must sort the result.
+func (s *Service) GetBySeller(sellerAgentURL string) []Entry {
+	snap := s.snap.Load()
+	entries := snap.bySeller[sellerAgentURL]
+	if len(entries) == 0 {
+		return nil
+	}
+	out := make([]Entry, len(entries))
+	copy(out, entries)
```
BaiyuScope3: nice to think about isolation here 👍 

But don't forget the `Entry` we're copying contains `*targeting.SegmentRule` which is still a pointer and copy(out, entries) performs a `shallow copy`. The test doesn’t catch this because it only mutates the Entry slot, not the rule inside it.

So if i appened something to the cfg.TargetSegments it might breaks behavior mentioned in the comments `mutating it does not affect the snapshot`

We could consider to do a deep copy or simply update the comments.

BaiyuScope3 (dismissed):

BaiyuScope3 (approved):

PR #132, opened 2026-05-20
refactor(identityagent): drop SHA-512 stubs and stub-ack flag

ohalushchak-exadel:
## Summary

- Remove SHA-512-truncation stub decoders (UID2 / EUID / PairID / PublisherFirstParty) and the operator-facing `TMPX_REFERENCE_STUB_ACK` acknowledgement that existed only to gate them.
- UID types without a registered decoder are now silently dropped from **both** the TMPX wire and the audience/fcap shadow request, driven by a single `Decode` pass and a single `Bytes == nil` chokepoint. `DecodedIdentity.HasReal` and `realTypes` map are gone.
- Delete `reference/identity-agent/` — `cmd/identity-agent/` is the production binary; the reference subtree was a duplicate. CI, AGENTS.md, go.work.example, and .gitignore files updated accordingly.
- Startup logging now categorises UID types into enabled / dropped / unreachable-in-priority, so a LiveRamp misconfiguration or a half-dead \`TMPX_PRIORITY\` list is visible at boot.
- Fix stale doc claim that ID5 was a stub — ID5 has been a real pass-through decoder since #128.

## Behavior changes for operators

- \`TMPX_REFERENCE_STUB_ACK\` is no longer accepted (and no longer required). Remove from any env files.
- With TMPX configured but no LiveRamp sidecar, every UID type other than MAID / HashedEmail / ID5 is silently dropped — this was already true for audience/fcap; it now also applies to the TMPX wire.
- New startup log lines under TMPX enable: \`enabled_uid_types\`, \`dropped_uid_types\`, and a \`WARN\` for any \`TMPX_PRIORITY\` entries that name a UID type without a configured decoder.
- The \`docs/network-surface.md\` env-var table previously listed reference-agent names (\`TMP_IDENTITY_TMPX_*\`, \`TMP_IDENTITY_REGISTRY_URL\`, etc.) that haven't worked since the reference agent was deleted. These now match the actual \`cmd/identity-agent\` env vars (\`TMPX_*\`, \`TMP_REGISTRY_URL\`, etc.).

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go vet ./...\` clean
- [x] \`go test -race -count=1 ./targeting/identityagent/... ./targeting/internal/tmpxdecoders/...\` green
- [x] \`go test -count=1 ./...\` green (root module)
- [x] \`go test\` green in \`cmd/identity-agent\`, \`reference/context-agent\`, \`bench\`, \`cmd/router\`, \`targeting/prommetrics\`, \`e2e\` submodules
- [x] New \`TestNoDecoder_DropsFromBothWireAndAudience\` locks in the wire+shadow drop contract in a single Decode pass


docs/network-surface.md
```diff
@@ -110,18 +110,18 @@ Exposure tracking uses encrypted TMPX tokens instead of a dedicated endpoint:
...
-| `--tmpx-encrypt-jwks-ttl` | JWKS poll interval (default 5 min — the spec's recommended cache TTL). |
-| `--tmpx-country` / `TMP_IDENTITY_TMPX_COUNTRY` | Country stamped into the TMPX plaintext header. |
-| `--tmpx-priority` / `TMP_IDENTITY_TMPX_PRIORITY` | Comma-separated UID type ordering used to truncate identities when the resolved set would exceed the 255-byte wire budget (e.g. `uid2,rampid,id5`). Without it, an over-budget set returns an error — the spec forbids arbitrary truncation. |
+| `TMPX_ENCRYPT_JWKS_URL` | Buyer's JWKS endpoint advertising the TMPX recipient (X25519, `adcp_use=tmpx-encrypt`, `alg=HPKE-DHKEM-X25519-HKDF-SHA256`). The agent polls this on `TMPX_ENCRYPT_JWKS_TTL` and picks the entry with the newest `iat` for sealing. |
+| `TMPX_ENCRYPT_JWKS_TTL` | JWKS poll interval (default 5 min — the spec's recommended cache TTL). |
+| `TMPX_COUNTRY` | Country stamped into the TMPX plaintext header. |
```
BaiyuScope3, line 119: Are we hard code this per deployment? thought it should be implied per request

ohalushchak-exadel: per spec, country is not forwarded from router to identity match. Can discuss/update separately

BaiyuScope3: then im curious, a deployment is per region, which could contains multiple countries, if we put country per deployment, it won't be accurate isn't it ?
It could be resolved from ip address or the context of a request

BaiyuScope3: Not a blocker at this point, more like a question

BaiyuScope3 (approved):

docs/network-surface.md
ohalushchak-exadel: will respond in DMs

PR #166, opened 2026-05-26
feat(identityagent): normalize seller_agent_url for fcap lookups

ohalushchak-exadel:
## Summary

- Identity-agent's fcap stage now reduces `req.SellerAgentURL` to its registrable domain via `urlutil.Registrable` before building `fcap.Field`, matching what frequency-writer applies when it writes markers. Any deviation in subdomain, scheme, or path between buyer and writer previously made markers unreachable and silently disabled caps.
- When the request URL is not reducible (raw IP, single-label host), the fcap stage skips its lookup — symmetric to frequency-writer's "skip on invalid URL" behavior — and records `StageFCap`/`error` so it's visible in metrics.
- `identityconfig.ResolveRequest` keeps the un-normalized URL because the config source keys on full URLs (e.g. `"https://seller.example.com/agent"`), so this PR scopes normalization to the marker-key path only.

## Why this is safe

- `urlutil.Registrable` is the exact helper frequency-writer uses (see `internal/processor/processor.go:139` in scope3data/frequency-writer), so both sides apply the identical transformation.
- Behavior change is limited to fcap field construction; audience and resolve stages are untouched.
- Existing tests that used the placeholder `"s"` as a seller URL have been updated to `"seller.com"` — `"s"` is not a registrable domain and would now be rejected by the new validation.

## Test plan

- [x] `go test ./targeting/identityagent/...`
- [x] `go test ./...`
- [ ] CI green
- [ ] Spot-check that a real seller URL like `https://api.adsrvr.org/path` now writes markers under `adsrvr.org` on both writer and reader sides

### DIFF
--- a/targeting/identityagent/service.go
+++ b/targeting/identityagent/service.go
@@ -3,6 +3,7 @@ package identityagent
 import (
 	"context"
 	"errors"
+	"log/slog"
 	"maps"
 	"sync"
 	"time"
@@ -12,6 +13,7 @@ import (
 	"github.com/adcontextprotocol/adcp-go/targeting/fcap"
 	"github.com/adcontextprotocol/adcp-go/targeting/identityconfig"
 	"github.com/adcontextprotocol/adcp-go/tmproto"
+	"github.com/adcontextprotocol/adcp-go/urlutil"
 )
 
 // Service composes the audience-only IdentityEngine with a frequency-cap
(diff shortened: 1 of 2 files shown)


ohalushchak-exadel:
Closing due to GitHub partial outage; reopening as a fresh PR.

PR #238, opened 2026-05-27
docs(adcp): document pricing threshold zero semantics

### LINKED ISSUE #225: Decide optional numeric policy for pricing floor and min-spend fields
Follow-up from #223 Argus review.

`PricingOption.FixedPrice` is now `*float64` because presence vs omission selects fixed-price vs auction semantics and explicit zero is valid at the schema level. Adjacent numeric fields still use plain scalars:

- `PricingOption.FloorPrice`
- `PricingOption.MinSpendPerPackage`

Argus noted these have the same wire-level omission-vs-zero question. Reconfirm the protocol semantics. If explicit zero is valid and distinct from omission, pointerize them with migration notes. If zero is semantically meaningless/invalid in practice despite `minimum: 0`, document the rationale or add an explicit lint waiver.

bokelley:
## Summary
- keep `PricingOption.FloorPrice` and `PricingOption.MinSpendPerPackage` as scalar thresholds
- document the semantics: zero and omission both mean no threshold has been declared
- add explicit optional-numeric scalar lint waivers with protocol rationale

Closes #225

## Schema check
Local 3.0.12 pricing schemas model both fields as optional numeric thresholds with `minimum: 0`; unlike `fixed_price`, neither field presence selects a oneOf branch.

## Verification
- cd adcp/schemas && python3 lint.py --strict
- cd adcp && go test ./...
- go test ./...

## Expert review
- Protocol reviewer is running; will amend if it flags a blocking issue.

### DIFF
--- a/adcp/schemas/lint.py
+++ b/adcp/schemas/lint.py
@@ -384,6 +384,12 @@ OPTIONAL_NUMERIC_SCALAR_OK = {
     # semantically identical to omission, and include the protocol rationale in
     # the reason string.
     # (type_name, json_name): reason
+    ('PricingOption', 'floor_price'): (
+        'zero and omission both mean no auction floor has been declared'
+    ),
+    ('PricingOption', 'min_spend_per_package'): (
+        'zero and omission both mean no minimum spend requirement has been declared'
+    ),
 }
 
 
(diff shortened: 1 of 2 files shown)


bokelley:
Closing rather than merging: protocol review found that the proposed zero-vs-omission rationale is not actually stated by the 3.0.12 pricing schemas. The schemas only say fixed_price presence selects fixed vs auction pricing; floor_price and min_spend_per_package are optional non-negative numerics without explicit omission semantics. Keeping #225 open for an upstream/schema-backed decision instead of encoding an inferred policy in Go.

PR #358, opened 2026-06-01
feat(registry): optional persistent Store with Valkey + Redis backends

ohalushchak-exadel:
## Summary

Adds an optional persistent **`Store`** for the registry indexes so a
process restart can resume the feed against populated memory instead of
an empty index. Comes with two interchangeable backends:

- **`registry/glidestore`** — `valkey-io/valkey-glide/go/v2`
- **`registry/redisstore`** — `redis/go-redis/v9`

Both backends use the same key layout, so either client can be pointed
at the same Valkey/Redis instance. Each lives in its own `go.mod` so the
lightweight `registry` module stays client-free for embedders.

### Library surface

- `registry/persist.go` — single `Store` interface covering cursor +
  property + agent + auth state. Embeds the existing `CursorStore`
  contract. All methods are idempotent. `ValidatePublisherDomain` is
  exported so backends can reject `|`, whitespace, and control bytes
  in `publisher_domain` — the field-name encoding `{domain}|{type}`
  depends on `|` being absent.
- `registry/memstore.go` — in-memory reference impl with a
  `var _ Store = (*MemoryStore)(nil)` compile-time assertion.
- `registry/{property,auth,agent}_index.go` — optional `WithStore(s)`
  hook. Mutators (`Put`/`Add`/`Remove`/`RemoveEntry`/`RemoveAgent`/`Clear`)
  now **return `error` and take `ctx`**. Memory is updated regardless;
  persist failures surface to the caller. Each index gets a
  `Hydrate(ctx)` for startup that's idempotent via an atomic flag and
  warns if mutated before being hydrated.
- `registry/syncer.go` — `Run` hydrates all three indexes **before**
  loading the cursor. A page that fails to persist any event does not
  advance the cursor — next `FetchFeed` re-delivers (Put/Remove are
  idempotent at both layers). On `cursor_expired`, the cursor is only
  wiped after every `Clear` on the persistent store succeeds, so
  pre-expiry revocations can never be lost. If `Hydrate` fails on a
  non-empty saved cursor, the cursor is dropped and a bootstrap is
  forced to avoid resuming past events with empty memory.

### Restart-safety guarantees

The original gap — saved cursor + empty index → silent data loss — is
closed:

- **Hydrate before cursor load** rebuilds memory from the Store.
- **Cursor advancement gated on persist** means a flaky backend can
  never produce a state where the cursor moved past events that were
  never persisted.
- **Cursor-expired Clear failures keep the doomed cursor** until the
  store is actually wiped.
- **Auth field-name encoding is validated** so a malformed feed event
  can't poison the hash keyspace via a `|` in `publisher_domain`.

### Idempotency

All writes are `HSET`/`HDEL`/`SET`/`DEL` against canonical hash/string
keys. Repeated `Put` replaces; `Remove` on a missing key is a no-op;
`Clear` on an empty namespace is a no-op. `ClearAuth` uses SCAN+DEL
which is not atomic — the Syncer is the only expected caller and
serialises against the feed loop.

### Optional

Indexes constructed via `NewPropertyIndex()` (no `.WithStore`) behave
exactly as before — no I/O, no error returns surface to existing
callers beyond a clean `nil`. The `ctx` parameter is the only signature
change, threaded through so a stuck Valkey can be cancelled.

## Test plan

Unit tests (`registry/`):
- [x] `TestSyncer_HydratesBeforeFeedLoop` — saved cursor + pre-seeded
      store → memory populated before feed loop starts.
- [x] `TestSyncer_CursorExpiredWipesPersistentStore` — expired cursor
      clears both memory and store, then re-bootstraps.
- [x] `TestSyncer_PersistFailureBlocksCursor` — transient persist
      failure keeps the cursor put; flipping the failure off lets the
      cursor advance on the next retry.
- [x] `TestSyncer_CursorExpiredClearFailureKeepsCursor` — failure in
      `Clear` during cursor-expired handling does NOT wipe the cursor.
- [x] `TestPropertyIndex_HydrateIsIdempotent` — second `Hydrate` is a
      no-op even if the store grew between calls.
- [x] `TestValidatePublisherDomain` — table covers `|`, whitespace,
      control bytes, empty, happy path.
- [x] Dual-write / hydrate / clear round-trips for all three indexes.

Integration tests (Valkey 9 testcontainer):
- [x] `registry/glidestore` — cursor, properties, agents, auth, key-
      prefix isolation (~3s).
- [x] `registry/redisstore` — same shapes (~3s).

Out of scope:
- [ ] `cmd/router/main.go` wiring — production turn-on is a follow-up
      PR. The library ships ready; nothing in the binary changes yet.

## Review history

- Bot review flagged auth-revoke cursor-advance + cursor-expired
  partial-Clear as must-fixes; both are now resolved by gating cursor
  advancement on persist success across every event type, not just
  auth.
- Independent review added the `|` encoding collision, dual-write
  aliasing (caller's pointer leaked into the Store), Hydrate race /
  idempotency, ctx propagation, and several smaller items. All
  addressed in commit `fbb0a38`.


registry/syncer.go
```diff
@@ -145,39 +200,60 @@ func (s *Syncer) Run(ctx context.Context) error {
...
-			return
+			return nil
 		}
 		if entry.AgentURL == "" || entry.PublisherDomain == "" {
 			s.log.Warn("authorization event missing required fields", "event_id", event.EventID)
-			return
+			return nil
 		}
-		s.auth.Add(entry)
+		return s.auth.Add(ctx, entry)
```
BaiyuScope3, line 272: If a invalid domain got populated , 
auth.Add -> store.PutAuth -> validatePublisherDomain will return "InvalidDomainError"
when the persistFailed = true, the same event will just re-delivers infinitely isn't it ?
should we drop the event so next event can processed?

ohalushchak-exadel: fixed

BaiyuScope3 (approved):

PR #367, opened 2026-06-07
tmproto: add TMPX receiver open API

bokelley:
Adds an exported TMPX receiver API that opens SealTmpx wire tokens with the buyer X25519 private key and returns the plaintext plus kid. Moves the HPKE open path out of test-only code, adds a private-key loader, and keeps the implementation aligned with the existing fixed TMPX HPKE suite. Hardens payload parsing with strict raw base64url decoding so non-canonical encodings are rejected. Adds round-trip, tamper, wrong-key, and non-canonical encoding tests.


bokelley:
Closing as duplicate of #366; this branch contains the same TMPX receiver API work.

PR #373, opened 2026-06-09
feat(reference): World ID verifier + runnable verified-identity agent

bokelley:
The concrete World ID verifier that makes the verified-identity seam (#370/#371) actually *do* something — plus a runnable reference buyer agent that proves the whole receiver path end to end on the shipped code.

## Why a new module
Own `go.mod` (mirroring `reference/seller-agent/`), so World's HTTP verify client never enters the **core** adcp-go dependency graph. That's the whole reason #370 made `targeting.AttestationVerifier` a pluggable interface with **no default** (nil ⇒ every attestation absent). Core `tmproto`/`targeting` untouched.

## `worldid.Verifier` (implements `targeting.AttestationVerifier`)
POSTs the attestation's proof to World's `verify/{rp_id}` (rp_id = the RP the SDK already bound via the audience), and derives the **nullifier + claims from World's authoritative response**. Verify-before-trust: the sender's asserted `att.Claims` are never read — only what World confirms enters the verified set. Any failure (no proof, non-2xx, `success:false`, missing nullifier, no recognised claim) → error → caller treats it as no attestation.

## Runnable agent (`main.go`)
Wires the verifier + an HPKE recipient key + a static age resolver into the `identityagent.Service`:
- `GET /recipient` → `{audience_kid, public_key, rp_id, seller_agent_url}` (the sender seals to this).
- `POST /seal` → demo helper: wraps a World ID proof in an attestation and HPKE-seals it (in production the network/publisher seals; the agent never does).
- `POST /identity` → an `IdentityMatchRequest` with `sealed_credentials`; runs the real eligibility pipeline.

## Tests
7 verifier unit tests against a mocked World backend + an **end-to-end test through the real `Service`** with real HPKE seal/open: World confirms only `unique_human`, so a sender-asserted `age_over_21` stays ungranted and the 21+ package is ineligible while the open package passes — verify-before-trust, demonstrated on shipped code. `go build` / `vet` / `test -race` all pass; the agent boots and serves.

## Deferred
World ID's age-credential identifiers (the claim map currently recognises `proof_of_human → unique_human`; age maps in as World exposes them), a production recipient-key/secret loader (env/generated for now), and the live browser run (wire the idkit widget → this agent).


bokelley:
Closing as superseded by #374, which merged the World ID verifier module and config-gated wiring into the deployable identity agent. The current tree has since evolved further, so this older parallel implementation is no longer a merge candidate.

PR #383, opened 2026-06-12
feat(targeting): verify in-band World ID attestations (Mechanism A)

BaiyuScope3:
## What

Adds verify-before-trust consumption of **in-band World ID attestations** (Mechanism A) — the `attestation` carried directly on `identity_match_request.identities[]` — alongside the existing `sealed_credentials` (Mechanism B) carrier.

## Why

rtdp forwards a network-scoped World ID proof on `Identities[].Attestation` (it relays seller-side and cannot HPKE-seal to an audience key). The identity agent previously consumed only `sealed_credentials`, so that attestation was never verified. This wires the in-band carrier through the same verifier seam.

## How

- **`targeting`**: extract `verifyOne` (shared `LocalPreCheck` → verifier → post-verify RP recheck → normalize); add `VerifyAttestations`, the no-HPKE sibling of `OpenAndVerify`.
- **`identityagent`**: the verified-identity stage runs both carriers and merges results, so frequency-cap `CapKey`, the verified-identity gate, and the TMPX nullifier encoding all light up. Adds `RelyingPartyID` config + `WithRelyingPartyID` option.
- **`cmd/identity-agent`**: recipient keys become optional (they power only the sealed carrier); `TMP_RELYING_PARTY_ID` drives the in-band carrier.

Fail-closed preserved: no verifier or no `TMP_RELYING_PARTY_ID` ⇒ no in-band verification, behavior unchanged. Claims are trusted only from the verifier, never the asserted attestation.

## Trust surface note

This makes the agent verify an in-band attestation against a single configured relying party (network-as-RP). That is a deliberate addition to the trust surface that #377 left out — flagging for reviewer attention.

## Tests

`go test ./targeting/...` green; `go build`/`go vet` clean on the root and the `cmd/identity-agent` module; `gofmt` clean.

### DIFF
--- a/cmd/identity-agent/verified_identity.go
+++ b/cmd/identity-agent/verified_identity.go
@@ -23,14 +23,18 @@ import (
 //
 // Required when enabled:
 //
-//	TMP_RELYING_PARTY_ID  the relying party this deployment acts as; must match
-//	                      the rp_id the sealed credentials were sealed for.
-//	TMP_RECIPIENT_KID     the audience_kid identifying our HPKE recipient key.
-//	TMP_RECIPIENT_KEY     hex-encoded 32-byte X25519 private key (secret —
-//	                      injected from secret storage, never logged).
+//	TMP_RELYING_PARTY_ID  the relying party this deployment acts as. In-band
+//	                      attestations on req.Identities must carry this rp_id,
+//	                      and sealed credentials must be sealed for it.
 //
 // Optional:
 //
+//	TMP_RECIPIENT_KID     the audience_kid identifying our HPKE recipient key.
+//	TMP_RECIPIENT_KEY     hex-encoded 32-byte X25519 private key (secret —
+//	                      injected from secret storage, never logged). KID and
+//	                      KEY are set together to enable the sealed_credentials
+//	                      carrier; without them only in-band attestations are
+//	                      verified.
 //	WORLD_VERIFY_BASE_URL World's verifier backend (defaults to production).
 //	WORLD_API_KEY         API key for World's verifier backend, if required.
 //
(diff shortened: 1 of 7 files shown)


EmmaLouise2018 (approved):

bokelley:
Reviewed — this is clean and well-scoped. The `verifyOne` extraction is the right move: both carriers (sealed / Mechanism B and in-band / Mechanism A) now run the *identical* `LocalPreCheck → verifier → post-verify RP recheck → normalizeClaimSet` sequence, so there's no verification drift between them. Fail-closed is preserved + tested, "claims only from the verifier" is tested, and recipient keys correctly become optional for an in-band-only deployment.

On the trust-surface point you flagged — agree it's worth calling out, and I'd sharpen what the delta actually is:

- **Verification security is unchanged.** Because A reuses `verifyOne`, forgery resistance (verifier checks the real proof) and replay resistance (`signal_binding` must commit to `vctx.RequestID`) are identical to the sealed path. An attestation observed on the wire can't be replayed to a *different* request.
- **The real delta is confidentiality.** The in-band attestation rides in the clear on the request (nullifier + claims visible to anyone who sees the `IdentityMatchRequest`). The seal never added authentication — sealing only needs the operator's public key — it added confidentiality + addressing. So Mechanism A is sound **only where the request channel is trusted end-to-end** (network-as-RP: the relay is the operator). Worth stating that boundary explicitly: the in-band attestation must not cross an untrusted hop (logging, forwarding, a third party) or the identity leaks.

So I'd reframe the note from "expands the trust surface" to "**drops confidentiality for the attestation; verification bar is unchanged**."

Two follow-ups before/at merge:

1. **Load-bearing dependency:** the unsealed path leans entirely on the verifier enforcing `signal_binding commits to RequestID` (the documented MUST in `verified_identity.go`). It's shared with the sealed path, so if B is safe A inherits it — but worth a one-line confirm that `verify/worldid` actually checks the *commitment*, not just non-empty.
2. **Nit:** if the same nullifier ever arrives via both carriers on one request, the merge appends both → potential double-count in the cap/vid gate. Unlikely in practice (a deployment uses one carrier), but a dedup-by-nullifier on the merged slice is cheap insurance.

Verdict: approvable. The only things I'd want are the confidentiality-boundary note in the docs and the `signal_binding→RequestID` confirmation.

bokelley:
Two items from the review applied in [43971b4](https://github.com/adcontextprotocol/adcp-go/commit/43971b4095446b671349c2572901935bc162e3de):

**1. Confidentiality boundary note (item 1 — applied)**
`VerifyAttestations` godoc now carries the boundary you described: no confidentiality envelope, delta vs sealed is confidentiality only, MUST NOT cross an untrusted hop. Framing matches your restatement — "drops confidentiality, verification bar unchanged."

**2. Dedup-by-nullifier (nit — applied)**
`runVerifiedIdentityStage` now deduplicates the merged slice by `CapKey()` (nullifier × rp) immediately after the carrier append, before the duration/outcome recording. Same-identity-both-carriers counts once. In-place filter; no extra allocation beyond the seen-map.

---

**3. `signal_binding → RequestID` commitment (item 2 — requires follow-up)**

The picture as it stands:

- `LocalPreCheck` (`targeting/verified_identity.go`) enforces `att.SignalBinding != ""` — non-empty, fail-closed ✓
- The `AttestationVerifier` interface contract (`verified_identity.go`) says `Verify` **MUST** confirm the signal_binding commits to `vctx.RequestID` ("the SDK guarantees signal_binding is non-empty but cannot check the scheme-specific hash")
- `verify/worldid.Verify` receives `vctx` (which carries `RequestID`) but only uses `vctx.ExpectedRelyingPartyID`. It marshals `att.Proof` directly to World's backend without passing or checking `att.SignalBinding` against a hash of `vctx.RequestID`

So: non-empty is checked; the commitment value is not. This gap pre-dates this PR and is shared with the sealed path — both run through the same `verifyOne` → `worldid.Verify`. Your "if B is safe, A inherits it" is accurate in terms of the delta, but neither carrier currently closes the commitment check.

I'd recommend tracking this as a follow-up security-sensitive issue against `verify/worldid` rather than blocking this PR on it. The fix would be for `worldid.Verify` to compute the expected signal binding from `vctx.RequestID` and confirm `att.SignalBinding` matches before (or after) the backend call. Happy to open that as a private issue if that's the right path.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #386, opened 2026-06-15
fix(tmp): make in-band World ID attestations reach the verified-identity stage (and stop the fcap panic)

BaiyuScope3:
Two related fixes that together let an in-band World ID attestation be accepted on the TMP identity-match path (and stop a crash). Found while debugging a World ID request that crash-looped the identity-agent and never produced a verified identity.

## 1. `identityagent`: attestations now reach the verified-identity stage under canonicalization
`buildServiceRequest` decodes inbound identities before `Evaluate`, but the canonicalizer drops `world_id_nullifier` (intentionally not inbound-decodable) and `audienceEligibleIdentities` rebuilds tokens **without** their `Attestation`. The verified-identity in-band path reads attestations off `req.Identities` *inside* `Evaluate`, so with a canonicalizer configured it saw no attestations → `inbandAvailable` false → **no verified identities ever produced**. World ID requests then reached the fcap stage with zero identities.

`serviceIdentities` now carries undecoded attestation-bearing identities (the World nullifier) through to the shadow request alongside the decoded audience-eligible set. `VerifyAttestations` skips attestation-less identities, fcap uses the verified `CapKey` branch once a verified identity exists, and the audience engine finds no segment match for a raw nullifier — so no existing path regresses.

## 2. `fcap`: `IsCappedAny` no longer returns a nil result that panics the caller
`IsCappedAny` returned `(nil, nil)` for empty `identities`, violating its documented "result has length `len(fields)`" contract. `runFcapStage` then indexed `cappedByField[i]` over the non-empty package set → `panic: index out of range [0] with length 0`, crashing the identity-agent into CrashLoopBackOff (failing every TMP request routed to the pod). It now returns an all-false result sized to `fields` (nothing can be capped without an identity), and `runFcapStage` defensively fails closed on any length mismatch instead of indexing out of range.

Observed in prod with image `identity-agent:main-cd3c9fa`: `panic ... runFcapStage ... index out of range [0] with length 0`. Tests updated/added for both fixes; full `./targeting/...` suite passes.

### DIFF
--- a/targeting/fcap/service.go
+++ b/targeting/fcap/service.go
@@ -107,16 +107,25 @@ func (s *Service) IsCapped(ctx context.Context, userIdentity string, field Field
 // scales with the result slice (one per call) rather than N×M.
 //
 // Both input slices are read-only and must not be retained past return.
-// Empty identities or empty fields returns (nil, nil) without touching the
-// store.
+// Empty fields returns (nil, nil) without touching the store. Empty identities
+// (with non-empty fields) returns an all-false result of length len(fields) —
+// nothing can be capped without an identity — preserving the length-len(fields)
+// contract callers index against.
 //
 // Goroutine-safe: each invocation uses its own scratch buffer and shares no
 // state with concurrent callers. The store call inherits the supplied ctx
 // and routes through the configured topology — cluster/shadow distribution
 // is unchanged from IsCappedBatch.
 func (s *Service) IsCappedAny(ctx context.Context, identities []string, fields []Field) ([]bool, error) {
-	if len(identities) == 0 || len(fields) == 0 {
+	if len(fields) == 0 {
 		return nil, nil
+	}
+	if len(identities) == 0 {
+		// No identity can be capped, so nothing is capped. Return an all-false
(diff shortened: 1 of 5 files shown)


EmmaLouise2018 (approved):

bokelley:
Post-merge follow-up from a quick review pass:

The fcap part looks right to me: `IsCappedAny` now preserves its len(fields) result contract for empty identity input, and the identity-agent length guard prevents this class from becoming a panic again.

One thing I would tighten in a follow-up: `serviceIdentities` currently re-adds any undecoded identity that carries an attestation, not just the World ID nullifier carrier this fix is aimed at. That means a structurally valid but unverified attestation can make an otherwise dropped raw token visible to audience/fcap lookup paths after canonicalization. Audience evaluation loops over every `req.Identities` token, and fcap fallback also uses every request identity when no verified identity is produced.

I would either pass these attestation carriers only to the verified-identity stage, or at least restrict the preservation path to `UIDTypeWorldIDNullifier` and only when in-band verification is configured. I would also add a regression test proving invalid/unverified attestation carriers do not become audience/fcap lookup identities.

PR #389, opened 2026-06-22
tmp: align validation, types_gen, and generator with AdCP 3.1 schemas

ohalushchak-exadel:
## Summary

Four conformance gaps where adcp-go's hand-written or generated wire shapes diverged from the pinned AdCP 3.1 TMP schemas:

1. **\`ValidateContextRequest\` required \`property_id\`** — the schema marks \`property_id\` optional and \`property_rid\` required. Swap the empty-check from \`property_id\` to \`property_rid\` so validation matches the schema (\`tmproto/validate.go\`).

2. **\`Offer.price\` was a non-pointer with \`omitempty\`** — \`omitempty\` never elides a non-pointer struct, so an absent price serialized as \`{\"amount\":0,\"currency\":\"\",\"model\":\"\"}\` and downstream strict validators with the \`model\` enum would reject it. Overlay change emits \`*OfferPrice\`; \`targeting.buildOffersFromContextConfig\` and a test fixture updated to the pointer shape (helper \`offerPricePtr\` collapses zero-value to \`nil\`).

3. **Response structs missed \`adcp_version\` / \`adcp_major_version\`** — both response schemas pull them in via \`allOf\` against \`core/version-envelope.json\`, but the generator silently ignored \`allOf\`. Add an opt-in \`-merge-schemas\` directory + \`overlay.merge_refs\` whitelist that merges referenced schema properties into the parent struct's field list. Whitelist only \`core/version-envelope.json\` — \`core/protocol-envelope.json\` carries transport-layer fields (status, context_id, …) that don't belong on the TMP wire body and is intentionally excluded.

4. **Missing \`ErrorCodeSellerNotAuthorized\` constant** — 3.1's \`tmp/error.json\` enum includes \`seller_not_authorized\` for sync-time provider rejection; the generated enum did not. Add it to the \`ErrorCode\` overlay.

## Side-effect changes

- **Router**: \`HandleContextMatch\` now enriches \`property_rid\` from the registry *before* validating. Publishers that send \`property_id\` alone keep working; the post-enrichment validate keeps the spec contract for downstream provider fan-out. (\`router/router.go\`)
- **tmpclient**: \`ActivateParams\` gains a \`PropertyRID\` field and \`Client.Activate\` passes it through, so callers can supply the wire-required identifier.
- **\`go fix ./...\` modernizations**: a handful of \`slices.Contains\`, \`slices.Sort\`, \`maps.Copy\`, and struct-tag-alignment changes in unrelated files.

## Test plan

- [x] \`go test ./...\` and \`go test ./internal/generate/...\` green
- [x] \`go vet ./...\` clean
- [x] \`go fix ./...\` applied
- [x] \`go generate ./tmproto/...\` reproduces the committed \`types_gen.go\` (no drift)
- [ ] Reviewer sanity-check the response field order under \`allOf\` merge — the envelope fields are prepended so \`AdcpVersion\` and \`AdcpMajorVersion\` render before per-task fields, matching the inlined pattern on the request schemas


tmproto/validate.go
```diff
@@ -119,13 +119,13 @@ func ValidateContextRequest(req *ContextMatchRequest) error {
 	if req.RequestID == "" {
 		return errors.New("request_id is required")
 	}
-	if req.PropertyID == "" {
-		return errors.New("property_id is required")
+	if req.PropertyRID == "" {
```
BaiyuScope3, line 122: matchesProperty (router/filter.go) is now out of sync with validation. Since validate.go requires property_rid and makes property_id optional, a request can arrive with only property_rid. But the slug allowlist at filter.go:98-108 still gates on property_id — matchGlob("pub-*", "") fails and we return false before the RID allowlist at line 113 is checked, so a slug-configured provider silently drops a request it should match.

Cleanest fix we could consider in the upstream: the router already resolves property_rid from property_id (router.go:161) — let’s also do the reverse via LookupByRID, filling whichever identifier the client omitted before matching. Then matchesProperty works as-is.

BaiyuScope3 (dismissed):

tmproto/validate.go
ohalushchak-exadel: updated

BaiyuScope3 (approved):

PR #390, opened 2026-06-24
router: close five conformance gaps vs the TMP router spec

ohalushchak-exadel:
## Summary

Five router-only conformance fixes vs \`docs/trusted-match/router-architecture.mdx\` and \`specification.mdx\`. Surfaced by a focused router-only audit; companion to the recently merged \`tmp: align validation, types_gen, and generator with AdCP 3.1 schemas\` work.

### #7 — Context Match merge: dedup by package_id + warn

Spec §"Response Aggregation": duplicate \`package_id\` across providers is a configuration error; the router MUST keep the first response received and SHOULD log a warning. Impl silently appended every provider's offers. \`mergeContextResponses\` now dedups by \`package_id\` and warns naming both providers when a duplicate is seen. The fan-out result type now carries per-provider attribution (\`contextResult{providerID, response}\`, parallel to the pre-existing \`identityResult\`).

### #2 — Identity Match merge: warn on duplicate package_id

Spec §"Response Aggregation" also requires a SHOULD-log on duplicates for identity match. Because providers omit ineligibility rather than declaring it ("yes-or-silent"), a duplicate manifests as the same \`package_id\` in multiple providers' eligible lists — track per-package which providers reported it and warn when the count exceeds one. Union eligibility is preserved (the spec's "must be in both" collapses to union when only yes-responses are observable on the wire).

### #3 — Accept schema-aligned config field names + YAML

Spec sample is YAML using \`provider_id\` / \`timeout_ms\` / \`priority\` per the \`ProviderRegistration\` schema. Impl was JSON-only with \`id\` / \`timeout\` (ns) and no \`priority\`, silently dropping a copy-pasted spec config to an empty provider list. \`ProviderConfig\` now accepts both name sets via a custom \`UnmarshalJSON\`; \`LoadServerConfig\` sniffs \`.yaml\`/\`.yml\` extensions and converts via \`yaml.v3\` before unmarshaling so all struct tags and custom unmarshalers are honored uniformly across formats. Existing impl-internal name configs continue to work — covered by a new \`serverconfig_test.go\`.

### #4 — \`/healthz\` route alongside \`/health\`

Spec §Monitoring advertises \`/healthz\` as the router health endpoint; impl exposed only \`/health\`. Register both — \`/healthz\` for k8s probes copied from the spec, \`/health\` kept for back-compat.

### #6 — Emit the documented Prometheus metric series

Spec §Monitoring names \`tmp_context_match_duration_ms\`, \`tmp_identity_match_duration_ms\`, \`tmp_provider_duration_ms\`, \`tmp_provider_timeout_total\`, \`tmp_provider_error_total\`, \`tmp_offers_total\` — none were emitted. Extend \`FanOutMetrics\` with per-provider hooks (\`ObserveProviderDuration\` / \`IncProviderTimeout\` / \`IncProviderError\` / \`AddOffers\`), wire them through \`fanOutContext\` and \`fanOutIdentity\`, register the request-duration histograms in \`cmd/router/main.go\`, and add \`CounterAdd\` to \`prommetrics\` so the offers counter can move in bulk. The pre-existing \`router_*\` series still emit so existing dashboards keep working alongside the new spec-aligned ones.

## Out of scope (intentionally)

Per the audit list, these stay open and will land separately:
- Publisher-to-router auth boundary (wire-breaking, but bigger design than this PR)
- TLS / HTTPS listener (deployment-shape question)
- Targeting signal namespacing (also has a spec-side gap)
- Discovery's empty-list semantics (spec ambiguity)
- Context signature cache key conservatism (cosmetic)

## Test plan

- [x] \`go test ./...\` and all sub-modules build + relevant tests green (\`tmproto\`, \`router\`, \`tmpclient\`, \`cmd/*\`, \`e2e\`)
- [x] \`go vet ./...\` clean
- [x] \`go fix ./...\` no further diff
- [x] New \`serverconfig_test.go\` covers JSON-alias and YAML config paths
- [ ] CI green


router/router.go
```diff
@@ -472,24 +601,60 @@ func mergeContextResponses(requestID string, responses []*tmproto.ContextMatchRe
...
 		if minServeWindowSec < 0 || resp.ServeWindowSec < minServeWindowSec {
 			minServeWindowSec = resp.ServeWindowSec
 		}
+		providerID := ""
+		if i < len(providerIDs) {
+			providerID = providerIDs[i]
+		}
 		for _, pkgID := range resp.EligiblePackageIDs {
 			eligibleSet[pkgID] = struct{}{}
+			pkgProviders[pkgID] = append(pkgProviders[pkgID], providerID)
```
bokelley, line 664: `pkgProviders[pkgID]` is appended unconditionally per ID, so a single provider that lists the same `package_id` twice in its own `eligible_package_ids` yields `pkgProviders[pkgID] == ["alpha","alpha"]` and trips the cross-provider `duplicate package_id across providers (configuration error)` warning with a duplicated provider name. The Context path deliberately avoids exactly this (it branches on `first == res.providerID` and emits a distinct within-provider message). `eligible_package_ids` arrives raw off the wire with no dedup, so this is reachable.

Log-only (eligibility is built from a set, so the merged response is correct) and untested — `TestMergeIdentityResponses_LogsDuplicateWarning` only uses two distinct providers. Suggest tracking *distinct* providers per `package_id` and mirroring the Context path's within-provider message.

router/router.go
```diff
@@ -317,29 +378,45 @@ func (r *Router) fanOutContext(ctx context.Context, providers []ProviderConfig,
...
+						ext.IncProviderError(p.ID)
+					}
+				}
 				return
 			}
 			if r.health != nil {
 				r.health.RecordSuccess(p.ID)
 			}
+			if ext := r.metricsExt(); ext != nil {
+				ext.ObserveProviderDuration(p.ID, float64(time.Since(callStart).Milliseconds()))
```
bokelley: `callStart` is captured before `callProvider`, but `ObserveProviderDuration` only runs on the success path — the timeout and error returns above exit first. So `tmp_provider_duration_ms` reflects only fast, successful legs and is blind to the slow-failure tail: a provider hanging to its deadline contributes a `tmp_provider_timeout_total` increment but no latency sample, biasing p95/p99 down.

The spec's §Monitoring table calls this "Per-provider response time" with no success qualifier, and this repo's own `healthcheck.go` times all outcomes before branching. Either observe `time.Since(callStart)` for every terminal outcome (mirroring healthcheck.go), or reword the help text at `cmd/router/main.go:84` to "…for successful calls." Same at the identity site (line 509).

router/router.go
```diff
@@ -317,29 +378,45 @@ func (r *Router) fanOutContext(ctx context.Context, providers []ProviderConfig,
...
 			if err := r.callProvider(callCtx, p.Endpoint+"/context", callBody, sigHeaders, &cmResp); err != nil {
+				timeout, parentCancelled := classifyCallFailure(callCtx)
 				if r.health != nil {
-					if callCtx.Err() != nil {
+					switch {
+					case timeout:
 						r.health.RecordTimeout(p.ID)
-					} else {
+					case !parentCancelled:
 						r.health.RecordFailure(p.ID)
```
bokelley, line 408: Worth calling out in the PR description as a deliberate behavior change: previously any non-timeout error recorded `RecordFailure`; now a parent-context cancellation (client disconnect / server drain) records **neither** a failure nor a timeout, and the `&& !parentCancelled` guard skips the ext metrics too. The new behavior is more correct — don't penalize a provider for the router's own cancellation — but there's no test for it. Suggest a test that cancels the parent context mid-call and asserts neither `RecordFailure`/`RecordTimeout` nor `IncProviderError`/`IncProviderTimeout` fires.

router/serverconfig.go
```diff
@@ -58,13 +63,25 @@ type ShutdownConfig struct {
...
+// internally so all struct tags and custom UnmarshalJSON implementations on
+// nested types (e.g. ProviderConfig schema-name aliases) are honored
+// uniformly across formats. Invalid providers are logged and skipped rather
+// than causing a hard error.
 func LoadServerConfig(path string) (*ServerConfig, error) {
 	data, err := os.ReadFile(path) //nolint:gosec // path is from CLI flag, not user input
 	if err != nil {
 		return nil, err
 	}
+	ext := strings.ToLower(filepath.Ext(path))
```
bokelley, line 103: The provider-field aliasing + YAML closes the gap for the `providers` array, but the documented `tmp-router.yaml` sample in `router-architecture.mdx` uses **top-level** keys that still don't map: `listen:` (struct wants `addr`), `health_check_interval_sec:` (struct nests it as `health_check.interval_seconds`), and `adaptive_timeout:` (no field). So a verbatim copy-paste of the spec sample still silently drops the **listen address** — the exact "silently dropping a copy-pasted spec config" failure this PR targets, one level up (the router then binds the empty/default addr).

`tls:` is acknowledged out-of-scope; `listen` isn't, and it's the dangerous one. Suggest aliasing `listen`→`addr` and `health_check_interval_sec`, or explicitly listing the remaining top-level mismatches as out-of-scope in the PR body.

router/config.go
```diff
@@ -52,6 +53,35 @@ type ProviderConfig struct {
 	AudienceKIDs []string `json:"audience_kids,omitempty"`
 
 	Timeout time.Duration `json:"timeout"`
+
+	// Priority is documented in the schema (tmp/provider-registration.json) and
+	// the spec config sample for future use (merge conflict resolution, adaptive
+	// timeout allocation). Accepted on the wire so spec-aligned configs parse,
+	// but has no effect on routing today.
+	Priority int `json:"priority,omitempty"`
```
bokelley, line 65: `Priority` is parsed here but never read in routing (`grep -rn '\.Priority' router/ cmd/` is empty). Per `provider-registration.json` this field *is* the conflict-resolution key: "the router keeps the offer from the higher-priority provider." But `mergeContextResponses` keeps the first-received offer instead (matching the contradicting prose in `router-architecture.mdx`). The spec contradicts itself here — filed as https://github.com/adcontextprotocol/adcp/issues/5722.

If we resolve toward the schema, sorting providers by `priority` before the dedup loop would make the field live, make the conflict winner deterministic, and satisfy the schema. Either way, worth a line in the PR noting `priority` is parsed-but-deferred.

bokelley (commented):
Reviewed against the actual TMP spec (`adcontextprotocol/adcp`: `docs/trusted-match/router-architecture.mdx`, `specification.mdx`, `provider-registration.json`). Overall this is solid and genuinely conformant:

- **All six metric names + types match the §Monitoring table verbatim** (the three `*_duration_ms` histograms and the `tmp_provider_timeout_total` / `tmp_provider_error_total` / `tmp_offers_total` counters). The spec even says "Alert on `tmp_provider_timeout_total` increasing," so the counter is required.
- **`/healthz` is the correct router endpoint** per the spec; the extra `/health` is harmless back-compat.
- **Context keep-first matches the `router-architecture.mdx` prose** ("keeps the first response received").
- **YAML + `provider_id`/`timeout_ms`/`priority` (defaults 50/0)** match the schema and the documented `tmp-router.yaml` sample.

The PR also surfaced **two spec-level inconsistencies**, filed separately (not blockers for this PR):
- Identity merge: intersection rule vs the silent-ignore rule → https://github.com/adcontextprotocol/adcp/issues/5721
- Context merge: "first response received" vs the `priority` field → https://github.com/adcontextprotocol/adcp/issues/5722

Inline notes below are all minor — none block. Nice tests (asserting on JSON log output and parsed config is the right boundary), and the `classifyCallFailure` timeout/parent-cancel/error split is a clean, careful piece of work.

router/serverconfig.go
```diff
@@ -19,6 +24,31 @@ type ServerConfig struct {
...
+	type serverConfigAlias ServerConfig
+	aux := struct {
+		*serverConfigAlias
+		Listen                  string `json:"listen"`
+		HealthCheckIntervalSec  *int   `json:"health_check_interval_sec"`
+	}{serverConfigAlias: (*serverConfigAlias)(c)}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	if aux.Listen != "" && c.Addr == "" {
```
bokelley, line 43: Minor inconsistency with both the comment and the sibling `ProviderConfig.UnmarshalJSON`. The doc comment says "Schema-aligned names take precedence when both appear," but `&& c.Addr == ""` makes `addr` (the impl name) win when both `addr` and `listen` are set — whereas `ProviderConfig` lets the spec name (`provider_id`) override `id` unconditionally. So the two unmarshalers now disagree on precedence, and the comment describes the opposite of what the code does.

Benign (only matters if a config sets both forms), but worth tidying: either drop the `&& c.Addr == ""` guard so `listen` wins (matching `provider_id`), or reword the comment to say the impl name wins when both appear. Non-blocking.

bokelley (approved):
Approving. The follow-up commit (`ab3781b`) cleanly resolves all five points from my earlier review — the within-provider-repeat split in `mergeIdentityResponses`, observing `tmp_provider_duration_ms` on all terminal outcomes except parent-cancel, the parent-cancel test, the top-level `listen` / `health_check_interval_sec` aliases, and deferring `priority` to adcontextprotocol/adcp#5722 — each with a matching test, and CI is green.

One non-blocking nit inline on the new `ServerConfig.UnmarshalJSON` precedence. The two spec contradictions this PR surfaced are tracked in adcontextprotocol/adcp#5721 (identity merge) and adcontextprotocol/adcp#5722 (context conflict resolution).

bokelley (approved):
Approved. I rechecked the router conformance patch and verified it locally with go test ./.... The changes are router-scoped, CI is green, and the remaining caveats are non-blocking follow-ups.

PR #391, opened 2026-06-25
targeting: direct-match fields on PackageContextConfig for buyer-typed value targeting

ohalushchak-exadel:
## Why

`PackageContextConfig` today expresses *all* value-based targeting (countries, regions, metros, content identifiers, language, sentiment, keywords, content policies) through `ContextSignals` cfgs + the `signal:*` keyspace. For buyer-typed value targeting where the publisher already sends the matched value on the request, this indirection is **tautological**:

- A `signal:42:country:US` key containing `"US"` in its CSV answers no question the request didn't already (the request says `geo.country=US`).
- Writers must coordinate two artifacts (a cfg in `context_signals.any_of` *and* a membership entry under a synthetic `signal_id`) to express what is fundamentally a direct value equality.
- For dimensions with no provider-defined membership (geo, content identifiers, language, sentiment) there is no "signal" entity — the indirection is overhead with no semantic content.

The signal mechanism remains correct for what it was designed for: third-party audience signals where membership is provider-defined and the lookup actually carries information. This PR doesn't remove or change that path. It adds direct-match fields for value-based dimensions so the signal path doesn't have to carry tautologies.

## What changed

**`targeting/entity.go`** — 13 new pairs of `[]string` fields on `PackageContextConfig` (inclusion + exclusion), plus a `MetroTarget` sub-struct for `{system, values[]}`:

- Geo: `Countries`, `Regions`, `Metros` + `*Exclude` (3 pairs)
- Content artifact identifiers: `EIDRs`, `Gracenotes`, `ISRCs`, `GTINs`, `RSSGUIDs`, `ISBNs` + `*Exclude` (6 pairs)
- ContextSignals projections: `Languages`, `Sentiments`, `Keywords`, `ContentPolicies` + `*Exclude` (4 pairs)

All fields use `omitempty`. Empty fields impose no constraint, so every existing `PackageContextConfig` keeps the same behavior.

**`targeting/engine_direct_match.go`** (new) — `directMatchPasses(cfg, req, country)`:

- Geo gates run country → region → metro, so a higher-level exclusion short-circuits before a more specific inclusion (per `docs/media-buy/advanced-topics/targeting.mdx`: *"exclusion at a higher level takes precedence over inclusion at a more specific level"*).
- Non-geo gates are independent: each runs in turn and the first failure rejects.
- Scalar fields use exact equality. List fields (`Keywords`, `ContentPolicies`, identifier lists) use intersection semantics: inclusion requires ≥1 overlap; exclusion rejects on any overlap.
- Identifier lists are matched against `ArtifactRefs` of the corresponding `Type` only — a `gtin` value cannot satisfy an `EIDRs` list.

**`targeting/engine.go`** — `directMatchPasses` is invoked between `matchesPropertyBitmap` and `signalsPass`, so it short-circuits before the signal MGet round-trip when a request fails a direct check.

**`targeting/metrics.go`** — `StageDirectMatch = "direct_match"` for evaluation telemetry.

## Tests

`targeting/engine_direct_match_test.go` covers:

- Per-field inclusion and exclusion paths.
- Inclusion rejects requests that don't supply the value at all (so a country-targeted package doesn't match a no-country request).
- Geo hierarchy resolution: country-exclusion beats region-inclusion; country-exclusion beats metro-inclusion; region-exclusion beats metro-inclusion.
- Metro `{system, value}` matching: system mismatch rejects (no system fanout).
- Content-identifier type filtering: a GTIN value on the request can't satisfy an `EIDRs` list.
- Independence of non-geo gates: any one gate's failure rejects.
- Backward compatibility: `PackageContextConfig{}` with no direct-match fields passes any request (the legacy behavior).

Full `go test ./...` is green; `go vet ./...` clean.

## Why this is implementation, not spec

The AdCP spec (`docs/trusted-match/specification.mdx`, conformance section) explicitly says: *"Storage backend (valkey, Aerospike, DynamoDB, in-memory, anything) is implementation. Two services with different storage backends that satisfy these invariants for the same inputs MUST return the same eligibility output."* The spec defines the wire protocol, eligibility invariants, and capability discovery — not the per-package config shape or matching algorithm internals. So this PR is purely an internal-storage / engine-logic addition; no spec change required.


bokelley (approved):
Approved. I reviewed the direct-match targeting change and verified it locally with go test ./.... The change is additive/internal to PackageContextConfig, keeps the TMP wire surface unchanged, and uses bounded evaluation telemetry.

PR #393, opened 2026-06-30
tmp-router: reproducible build pipeline for TEE attestation

ohalushchak-exadel:
Bucket 2 of the TEE work. Lands the build-side of the chain; the wire spec is in [adcontextprotocol/adcp#5770](https://github.com/adcontextprotocol/adcp/pull/5770) and the in-enclave runtime work (bucket 3) is a separate PR.

## Why this exists

A TEE attestation document carries a measurement of the running workload (Nitro PCR0, TDX MRTD, SEV-SNP SNP_MEASUREMENT, GCP Confidential Space image digest). A verifier compares that measurement against an allowlist. The allowlist is only as trustworthy as the procedure that produces it — anyone allowlisting a digest needs to be able to rebuild the same source and recompute it. Today our `cmd/router` build is not deterministic; this PR fixes that.

## What changes

| File | Change |
|---|---|
| `cmd/router/Dockerfile` | Base images pinned by multi-arch index digest (renovate-annotated). Go build adds `-buildvcs=false` and `-ldflags="-buildid="` to strip path, VCS, and build-id entropy. `SOURCE_DATE_EPOCH` wired through as a build-arg so BuildKit normalizes layer mtimes. |
| `scripts/build-tmp-router.sh` | NEW. Local reproducible-build script. Derives `SOURCE_DATE_EPOCH` from the last commit that touched a build input; calls `docker buildx` with the same flags CI uses; prints the OCI image digest; optionally emits a `tmp-router-measurements/v1` JSON manifest. |
| `.github/workflows/tmp-router-image.yml` | NEW. Parallel to `context-agent-image.yml`. Builds `linux/amd64` + `linux/arm64`, pushes to `ghcr.io/<owner>/<repo>/tmp-router`, signs each pushed digest with cosign keyless OIDC, writes the measurements manifest as a workflow artifact, and on tag pushes attaches it to the published image as a cosign attestation so verifiers discover it from the registry. |
| `docs/tmp-router-reproducible-build.md` | NEW. Operator/auditor verification procedure, what is and is not pinned, the OCI-digest → per-platform-measurement mapping table, and what to do on a reproducibility divergence. |

## Verified locally

Ran `scripts/build-tmp-router.sh` twice back-to-back on this branch; both runs produced the same OCI image digest:

```
sha256:40fd2ce72ee3e5c9bacf2651e64f925a4d95780b0bec28ef8a028b763da9909e
```

The script also produces a `tmp-router-measurements/v1` manifest with the digest, the source revision, and `dirty: true/false` based on the working tree state.

## How verifiers will use it

| TEE format | Measurement | Relation to published OCI digest |
|---|---|---|
| GCP Confidential Space | Workload-image digest in the token's `submods.confidential_space.image_digest` claim | **Direct equality** |
| AWS Nitro Enclaves | PCR0/PCR1/PCR2 in the attestation document | Derived deterministically from the OCI image plus the Nitro CLI version on a Nitro-enabled host |
| Intel TDX | MRTD in the quote | Same derivation model, with the TDX measurement-build chain |
| AMD SEV-SNP | SNP_MEASUREMENT in the report | Same derivation model |

For non-GCP cases the operator runs a one-shot transformation from this OCI image to the platform-specific measurement on a host with the relevant tooling — out of scope for this PR (GitHub-hosted runners don't have Nitro / TDX / SEV-SNP CLIs).

## Out of scope (explicit non-goals)

- Per-platform PCR / MRTD / SNP_MEASUREMENT derivation and publication. That happens on a controlled host with the platform tooling, separately per release.
- Transparency-log of measurements beyond what Sigstore already provides via the cosign signatures and the `provenance: mode=max` in-toto attestation.
- In-enclave key custody, KMS binding, attestation endpoint — bucket 3, separate PR.

## Test plan

- [x] `scripts/build-tmp-router.sh` runs cleanly, prints the OCI image digest.
- [x] Two consecutive local runs on the same source produce **byte-identical** image digests.
- [x] `scripts/build-tmp-router.sh --measurements-out path.json` writes a valid `tmp-router-measurements/v1` manifest.
- [x] `.github/workflows/tmp-router-image.yml` parses as valid YAML.
- [x] `bash -n scripts/build-tmp-router.sh` passes.
- [ ] CI workflow runs successfully on this PR (build-only, no push, on `pull_request` trigger).
- [ ] Reviewer confirms the OCI-digest → per-platform-measurement mapping is the right factoring (vs. attempting platform-specific builds in this same CI).
- [ ] On first non-PR merge to `main` or first tag push, the GHCR package gets created (it'll be private by default; a repo admin flips it to public per the workflow header comment).


bokelley:
Closing because the protocol anchor adcontextprotocol/adcp#5770 was closed without merge, and this branch now conflicts with main. Reopen or replace this once the attestation contract is ratified; TEE measurement and publication machinery should not land ahead of its governing protocol.

PR #400, opened 2026-07-15
tmproto: agent-aware keystore for per-agent lazy key resolution

ohalushchak-exadel:
## Motivation

Today, TMP signature verification requires the verifier to sync every publisher's signing keys ahead of time — `RemoteKeyStore` polls a single snapshot URL that must contain every kid the verifier might ever need. That model fits routers well (they know which properties they sign for at startup), but not identity/context agents:

- The set of routers a provider will see is not known ahead of time.
- Providers typically see traffic from a small fraction of registered routers.
- Every startup pays the cost of fetching keys that will never be used.
- There is no widely deployed \"bulk snapshot\" endpoint today. `agenticadvertising.org` does not currently emit `{properties[]{signing_keys[]}}` — its equivalent is `/api/registry/authorizations` (per-agent narrow pull) and `/api/registry/authorizations/snapshot` (NDJSON), both keyed by `agent_url`.

The wire already carries the missing context. Every TMP request has `seller_agent_url` as a required signed field (specification.mdx §514, §555, §601-603 — the receiver \"uses this to resolve the publisher's signing keys for verification\"). All the verifier needs is for that URL to reach the keystore.

## Changes

- **New `AgentAwareKeyStore` interface** extending `KeyStore` with `LookupKeyForAgent(kid, sellerAgentURL string)`. Additive — existing implementations (`StaticKeyStore`, `RemoteKeyStore`) work unchanged.
- **`resolveSigningKey`** now takes `sellerAgentURL` and prefers the richer interface when the store implements it. Falls back to `LookupKey(kid)` otherwise.
- **`VerifyContextMatch` / `VerifyIdentityMatch`** pass `req.SellerAgentURL` through. Public function signatures unchanged.
- **New `LazyAuthorizationKeyStore`** — fetches per-agent from the AdCP registry `/api/registry/authorizations?agent_url=<canonical>` endpoint:
  - Per-agent cache keyed by `urlcanon.Canonicalize`'d URL so lookups are stable regardless of caller form.
  - Positive TTL 5 min (spec §579), negative TTL 60 s to absorb bursts of forged requests to unknown agents.
  - Single-flight collapses concurrent misses on the same URL onto one HTTP call.
  - `Invalidate(agentURL)` drops the entry so a verifier can re-fetch on signature failure (spec §590 requires this).
  - Bearer-token auth. Cross-origin redirects denied.

## Tests

10 new tests in `tmproto/keystore_authorization_test.go` cover fetch/cache, unknown-kid, 404 negative cache, TTL expiration, Bearer header, URL canonicalization, single-flight, Invalidate, https-only guard, and end-to-end `VerifyContextMatch` against a locally-signed request. Existing tmproto tests unchanged.

## Deployment shape (illustrative — no code here)

```go
ks, _ := tmproto.NewLazyAuthorizationKeyStore(tmproto.LazyAuthorizationKeyStoreOptions{
    BaseURL:     "https://agenticadvertising.org/api/registry/authorizations",
    BearerToken: os.Getenv("ADCP_REGISTRY_TOKEN"),
})
mux.Handle("/identity", tmproto.VerifyIdentityMatchHandler(handler, tmproto.VerifyOptions{
    OwnEndpointURL: "https://tmp.example.com/identity",
    KeyStore:       ks,
}))
```

## Related spec sections

- §506-518: signing model, seller_agent_url binding.
- §571-577: signature verification; providers use the publisher's public key from the registry.
- §579-593: key rotation and revocation — refetch on verify failure. Matches `Invalidate` semantics here.
- §601-603: key distribution via the property registry.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,6 +39,7 @@ The targeting engine (`targeting/`) is the shared evaluation core. Reference age
 | `tmproto/signing.go` | TMP request-authentication envelope (Ed25519, `X-AdCP-Signature`/`X-AdCP-Key-Id`, JCS for identity match, daily-epoch replay window). |
 | `tmproto/verify_middleware.go` | `VerifyContextMatchHandler` / `VerifyIdentityMatchHandler` middleware used by reference providers. |
 | `tmproto/keystore_remote.go` | `RemoteKeyStore` polls the router's `/registry/snapshot` for signing keys. |
+| `tmproto/keystore_authorization.go` | `LazyAuthorizationKeyStore` fetches signing keys on demand per `seller_agent_url` from the AdCP registry `/api/registry/authorizations` endpoint. Implements `AgentAwareKeyStore`. |
 | `router/serverconfig.go` | Config loading (JSON file, env vars, defaults). |
 | `cmd/router/main.go` | Router binary entry point — wires components, Prometheus metrics, env vars. |
 | `docs/network-surface.md` | Port map, data flow, pinhole spec, env var reference. |
--- a/targeting/contextagent/config.go
+++ b/targeting/contextagent/config.go
@@ -84,12 +84,22 @@ type Config struct {
 	Pprof   PprofConfig
 }
 
-// TMPConfig drives TMP signature verification on /context.
+// TMPConfig drives TMP signature verification on /context. See
+// identityagent.TMPConfig for full field semantics — the two are kept in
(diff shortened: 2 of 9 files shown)


ohalushchak-exadel:
Reopening from the base repo so CI has access to secrets (the AI-review job requires an app-id that isn't exposed to fork PRs). New PR: same commits pushed to `agent-aware-keystore` in this repo.

PR #404, opened 2026-07-16
identityagent: implement multi-chunk TMPX sealing

ohalushchak-exadel:
## Motivation

`targeting/identityagent/tmpx.go` previously carried a `warnIfMultiSlotIgnored` helper and a `MacroEntry` implementation that filled only the first configured macro slot, with an explicit warning that "multi-chunk encoding is not implemented; only the first slot will be filled on responses". The AdCP spec's `TmpxMacros` field allows up to two chunks in v1 (see `tmproto/types_gen.go` — `IdentityMatchResponse.TmpxMacros` and `ProviderRegistration.TmpxMacros`), so a provider registering `PIN_TMPX_1, PIN_TMPX_2` could not actually take advantage of the doubled ad-server slot budget. This PR closes that gap.

## What changes

- `TMPXSealer.wireBudget()` — new helper returning `max(1, len(macroNames)) * TmpxMaxWireBytes`. Each configured macro slot contributes one GAM `%%PATTERN_MACRO%%`-sized chunk to the wire budget; no-macro deployments keep the legacy single-slot budget.
- `TMPXSealer.selectEntries` — now packs entries against `wireBudget()` instead of a hard-coded `TmpxMaxWireBytes`. The existing priority-truncation contract is preserved: an overflow of the multi-slot budget still drops from the tail of `TMPX_PRIORITY` rather than erroring, and priority-empty deployments still fail closed.
- `TMPXSealer.MacroEntries(token)` — replaces the previous single-entry `MacroEntry`. Splits the sealed wire token at 255-byte boundaries in slot order, emits one `TmpxMacro` per non-empty chunk, and never emits more entries than `macroNames` or a chunk larger than one slot.
- `handler.go` — extracts the tmpx-response wiring into `assignTmpxToResponse`. `resp.TmpxMacros` now carries the chunked entries; the deprecated `resp.Tmpx` single-string field is populated only when the sealed token fits within one slot (a multi-chunk value cannot be represented in a single 255-byte carrier).
- Removes the now-obsolete `warnIfMultiSlotIgnored` helper and its startup warning.

## Test coverage

- Single-slot passthrough: `TestMacroEntries_SingleSlotEmitsWholeToken`, `TestSelectEntries_NoPriorityPassesUnderBudget` (unchanged).
- Multi-slot, fits in one slot: `TestMacroEntries_TwoSlots_TokenFitsInFirst` — only slot 1 is emitted; the trailing empty slot is dropped.
- Multi-slot, straddles both slots: `TestMacroEntries_TwoSlots_TokenStraddlesBothSlots` — chunk boundary at exactly 255 bytes, reassembly by byte-concatenation.
- Multi-slot, priority-truncation on overflow: `TestSelectEntries_TwoSlotBudget_FitsMoreEntries`, `TestSelectEntries_TwoSlotBudget_PriorityTruncatesOnOverflow`.
- End-to-end seal + chunk: `TestSeal_TwoSlotEndToEnd` — a real HPKE seal produces a wire > 255 bytes and `MacroEntries` reassembles it losslessly.
- Handler wiring: `TestAssignTmpxToResponse_SingleSlotFillsBothCarriers`, `TestAssignTmpxToResponse_MultiChunkOmitsLegacyTmpx`, `TestAssignTmpxToResponse_NoMacrosKeepsOnlyLegacyTmpx`.

## Backward compatibility

Single-slot operators — the only production shape this codebase has emitted so far — are unaffected: `wireBudget()` returns `TmpxMaxWireBytes`, `selectEntries` behavior is unchanged, `MacroEntries` returns one entry carrying the whole token, and the legacy `resp.Tmpx` field is still populated. No-macro operators are also unchanged.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -660,7 +660,17 @@ func mergeIdentityResponses(requestID string, providerIDs []string, responses []
 			tmpxProviders[providerID] = tmproto.TmpxProviderEntry{
 				Macros: append([]tmproto.TmpxMacro(nil), resp.TmpxMacros...),
 			}
-			if legacyTmpx == "" {
+			// The deprecated single-string `tmpx` carrier can only represent
+			// tokens that fit in one macro slot. Multi-chunk responses (>1
+			// TmpxMacros entry) produce a wire that spans multiple slots —
+			// any single chunk on its own is a broken ciphertext half that
+			// fails AEAD open silently. Populate the legacy field only when
+			// the emitting agent produced exactly one chunk; a multi-chunk
+			// emission leaves it empty so consumers still reading the
+			// deprecated field skip it rather than get identity loss. See
+			// the mirror guard in identityagent/handler.go's
+			// assignTmpxToResponse.
+			if legacyTmpx == "" && len(resp.TmpxMacros) == 1 {
 				legacyTmpx = resp.TmpxMacros[0].Value
 			}
 		} else if resp.Tmpx != "" {
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -206,6 +206,38 @@ func TestMergeIdentityResponses_TmpxProvidersFromNewShape(t *testing.T) {
 		"legacy tmpx must mirror the first provider's first slot for back-compat")
 }
 
+// TestMergeIdentityResponses_MultiChunkClearsLegacyTmpx pins the guard
+// against the identity-loss regression Blocker 2 fixed: when an agent emits a
+// multi-chunk sealed token (len(TmpxMacros) > 1), the deprecated single-string
+// `tmpx` field cannot carry it — any one chunk on its own is a broken
+// ciphertext half that fails AEAD open silently. The router MUST leave the
+// legacy carrier empty in this case so consumers still reading `tmpx` skip it
+// rather than receive garbage; the authoritative multi-chunk data survives
+// intact on TmpxProviders[provider_id].Macros.
+func TestMergeIdentityResponses_MultiChunkClearsLegacyTmpx(t *testing.T) {
+	multi := &tmproto.IdentityMatchResponse{
+		EligiblePackageIDs: []string{"pkg-1"},
+		ServeWindowSec:     300,
+		TmpxMacros: []tmproto.TmpxMacro{
+			{Name: "PIN_TMPX_1", Value: "k1.first-half-of-sealed-token"},
+			{Name: "PIN_TMPX_2", Value: "second-half-of-sealed-token"},
+		},
+	}
+
+	merged := mergeIdentityResponses("id-multi", []string{"pinnacle_id"},
+		[]*tmproto.IdentityMatchResponse{multi}, nil)
+
+	require.NotNil(t, merged.TmpxProviders, "tmpx_providers MUST carry the multi-chunk data")
+	pin, ok := merged.TmpxProviders["pinnacle_id"]
+	require.True(t, ok, "pinnacle_id entry must exist")
+	require.Len(t, pin.Macros, 2, "both chunks MUST survive on tmpx_providers")
+	assert.Equal(t, "k1.first-half-of-sealed-token", pin.Macros[0].Value)
+	assert.Equal(t, "second-half-of-sealed-token", pin.Macros[1].Value)
+
+	assert.Empty(t, merged.Tmpx,
+		"multi-chunk emissions MUST leave the deprecated legacy `tmpx` field empty — a single chunk is a broken half-token")
+}
+
 // TestMergeIdentityResponses_LegacyOnlyAgent covers a transition case: an
 // agent that only emits the deprecated `tmpx` string with no TmpxMacros[].
 // The router preserves it in the legacy field but does NOT synthesize a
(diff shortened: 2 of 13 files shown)


ohalushchak-exadel:
Addressed all three MUST-FIX blockers and the MacroEntries follow-up. Pushed to `multi-chunk-tmpx` as four focused commits.

**Blocker 1** — `OpenTmpx` refused reassembled multi-chunk tokens (`tmproto/tmpx_open.go:70`)

Introduced `TmpxMaxSlots` (v1 cap = 2) and `TmpxMaxReassembledWireBytes = TmpxMaxSlots * TmpxMaxWireBytes` in `tmproto/tmpx.go`. `OpenTmpx` and `TmpxKid` now bound against the reassembled budget (510 bytes) rather than the per-slot substitution limit (255). `OpenSealedCredential` is unchanged — it retains its distinct, larger sealed-credential bound. Added `TestOpenTmpxAcceptsReassembledMultiChunk` in `tmproto/tmpx_sealed_test.go` that seals a >255-byte wire, splits at the slot boundary as `MacroEntries` would, and pins byte-concatenation round-trip. `TestOpenRejectsOversizedWireBeforeDecode` and `TestOpenSealedCredentialLargerBound` are resized to the new bound.

**Blocker 2** — Router truncated legacy `tmpx` for multi-chunk responses (`router/router.go:664`)

The legacy single-string carrier now populates only when `len(resp.TmpxMacros) == 1`. Multi-chunk emissions leave `merged.Tmpx` empty; the authoritative data survives on `merged.TmpxProviders[provider_id].Macros`. Mirrors the sender-side guard in `assignTmpxToResponse`. New `TestMergeIdentityResponses_MultiChunkClearsLegacyTmpx` pins the invariant.

**Blocker 3** — No runtime cap on `TMPX_MACRO_NAMES` (`parseTmpxMacroNames` in `targeting/identityagent/config.go`)

`parseTmpxMacroNames` now returns `([]string, error)` and rejects any list longer than `tmproto.TmpxMaxSlots` at startup with a message naming the cap, the source of truth (`provider-registration.json tmpx_macros.maxItems`), and the receiver bound the cap enforces. `TestLoadConfigFromEnv_TmpxMacroNames` covers empty, one, two, and three-entry cases.

**Follow-up** — `MacroEntries` silent surplus-drop (`targeting/identityagent/tmpx.go:272-274`)

`MacroEntries` now returns nil when the token exceeds `len(macroNames) * TmpxMaxWireBytes`, surfacing the condition as "no TMPX emitted" rather than emitting truncated chunks the receiver cannot reassemble. `TestMacroEntries_FailsClosedOnOverBudgetToken` (formerly `_ClampsToConfiguredSlotCount`) pins the fail-closed shape.

`go build`, `go vet`, and `go test ./...` all clean. Ready for another look.

bokelley (approved):
Ran `go build ./...`, `go vet ./...`, and `go test ./...` on the four follow-up commits — all 25 packages green. Reviewed the diff against the three blockers and the MacroEntries follow-up.

**Blocker 1 — OpenTmpx bound** ✓  
`TmpxMaxSlots = 2` and `TmpxMaxReassembledWireBytes = 510` are correctly placed in `tmproto/tmpx.go`. `OpenTmpx` and `TmpxKid` now gate on the reassembled budget; `OpenSealedCredential` is untouched. `TestOpenTmpxAcceptsReassembledMultiChunk` correctly simulates receiver-side reassembly (split at 255, rejoin, call `OpenTmpx`) and the round-trip holds. `TestOpenRejectsOversizedWireBeforeDecode` is correctly re-anchored to the new ceiling — the old 255-byte ceiling would now reject legitimate multi-chunk output.

**Blocker 2 — Router legacy field** ✓  
`len(resp.TmpxMacros) == 1` guard is the right predicate. Single-chunk agents keep back-compat; multi-chunk agents leave `merged.Tmpx` empty so a consumer reading the deprecated field gets a clean skip rather than a half-token AEAD failure. `TestMergeIdentityResponses_MultiChunkClearsLegacyTmpx` pins this.

**Blocker 3 — TMPX_MACRO_NAMES cap** ✓  
`parseTmpxMacroNames` → `([]string, error)` threaded cleanly through `LoadConfigFromEnv`. The error message names the cap, the source of truth (`provider-registration.json`), and the receiver bound. The four-case test (empty / 1 / 2 / 3) is complete.

**MacroEntries follow-up** ✓  
`MacroEntries` returns nil on over-budget token instead of truncating. The sealer's `selectEntries` keeps tokens within `wireBudget()` so this path is a defensive floor — correct shape for that invariant.

**One nit (non-blocking):** `TMPXSealer.macroNames` field comment still says "the sealer does not impose an extra cap so an operator can experiment beyond v1 without this code needing a bump." That was written before Blocker 3 landed `parseTmpxMacroNames` with a hard cap. The sealer struct itself is correctly permissive (cap lives at the config layer), but the comment now slightly misrepresents the effective system behaviour. Worth a one-line update in a follow-up to avoid confusion for the next reader.

Overall: all three blockers and the follow-up are correctly addressed. Code is clean, tests are thorough, and no regressions. LGTM.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #2, opened 2026-03-26
Publish v0.1.0 as a proper Go module


bokelley:
Remove replace directives, tag v0.1.0, verify go get works. Consider restructuring as single module with cmd/ for binaries.

PR #406, opened 2026-07-16
trusted-match: bump adcp pin to 3.1.3 and close identity/context spec gaps

ohalushchak-exadel:
## Summary

- Bumps the AdCP schema pin from 3.1.1 → 3.1.3 (the latest released version). No trusted-match schema changes in that range — the delta is media-buy: `ProductFilters.publisher_domain` (new) and delivery-metrics `CompletionRate *float64` (nullable).
- Rebased onto latest `main`.
- Fixes three trusted-match spec gaps surfaced by a fresh audit:
  1. **identity-match `seller_agent_url` canonicalization** — the request URL was being compared byte-exact against identityconfig snapshot keys; context-match already canonicalized, identity-match did not. A publisher sending `https://Seller.Example.com:443/agent` against a registration under `https://seller.example.com/agent` returned an empty eligible set. Fixed on both the read path (`identityagent`) and the storage layer (`identityconfig` snapshot builds).
  2. **context-match ladder validation** — `ContextSignals.Validate()`, `ArtifactRef.Validate()`, and `Artifact.Validate()` existed but were never called from `ValidateContextRequest`, so requests with `sentiment: "foo"`, 60 topics, or artifacts missing `property_rid` were silently accepted. Now enforced at handler entry. Also adds `validateGeo()` for the country / region patterns and `metro.{system,value}` required subfields the JSON Schema encodes on `Geo map[string]any`.
  3. **identity-match consent enforcement** — the spec's prose "buyers in regulated jurisdictions MUST NOT process the user token without consent information" is now a hard check: when `consent.gdpr == true`, either `tcf_consent` or `gpp` must be non-empty.

## Not fixed here (follow-ups)

- **`seller_not_authorized` at sync time** — the audit called this out as unimplemented. adcp-go doesn't currently ingest `AvailablePackage`s from a sync flow anywhere in the tree, so there's no callsite to hook the check into. Belongs to whoever adds the sync-side integration.
- **Nullifier-reuse tracking** — self-documented at `targeting/verified_identity.go:79`; a separate design step.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite passes; one router test updated to include the now-required `Artifact.property_rid` + `artifact_id`)
- [x] `adcp/schemas/check-freshness.sh` — pinned + no schema drift (reports 3.2.0 as newer per the CDN manifest, but that version has no GitHub release yet)
- [x] `golangci-lint run` — no new issues; pre-existing gosec warnings in untouched files remain
- [x] New regression tests: `TestResolveRequest_CanonicalizesSellerAgentURL`, `TestValidateContextRequest_LadderConstraints`, `TestValidateIdentityRequest_Consent`


ohalushchak-exadel:
On the blocking question — schema types \`consent.gdpr\` as \`boolean\`, not IAB's integer \`0/1\`:

https://github.com/adcontextprotocol/adcp/blob/v3.1.3/static/schemas/source/trusted-match/identity-match-request.json#L131-L149

\`\`\`json
"gdpr": {
  "type": "boolean",
  "description": "Whether GDPR applies to this request."
}
\`\`\`

AdCP is intentionally distinct from OpenRTB here — the \`.(bool)\` assertion in \`validateConsent\` matches the wire contract. A publisher sending \`{gdpr: 1}\` is off-spec; rejecting it is the correct behavior.

Pushed a small follow-up (13a0ed4) for the two non-blocking comment nits (source-path in \`validate_ladder.go\`, double-canonicalization note in \`identityagent.Evaluate\`). Filing the fcap-key-asymmetry as a tracked issue next.

ohalushchak-exadel:
Addressed the fcap namespace split in f56ce27.

**What changed**
- `runFcapStage` now receives the canonicalized seller URL from `Evaluate` and builds `fcap.Field` with it. The old "seller URL is the marker key verbatim" invariant is retired; the new invariant is symmetric canonicalization on both sides. Comment on the stage now documents the writer-side requirement.
- Canonicalization is best-effort: canonical form for well-formed http/https URLs, raw pass-through for non-URL routing strings (e.g. the storefront paths added in commit 0bee186). Applied consistently in `identityagent.canonicalizeSellerAgentURL` and `identityconfig.canonicalSellerKey` so read and write paths stay in lockstep on both spec-compliant URLs and internal routing forms. Dead nil-guard branches in `identityconfig.{Get,Lookup,GetBySeller,canonicalizeEntry}` dropped.
- Regression tests: `TestService_FCap_CanonicalizesURLVariants` (mixed case / default port / dot-segment all hit the same cap bucket) and `TestService_FCap_StorefrontPathPassesThrough` (path-only routing string still works byte-for-byte). The old `TestService_FCap_UsesSellerURLVerbatim` was passing vacuously post-#406 initial commits — replaced with tests that actually pin the new invariant.

**Downstream writer coordination**

Any service that constructs `fcap.Field` directly to record markers must canonicalize with the same helper or cap keys fragment by URL spelling. Details, one-line fix, rollout considerations, and a proposal to promote canonicalization into the `fcap` package as a supported API are tracked in issue #408.

**On MUST-FIX #2 (breaking-change marker for `completion_rate`)**

That one I'd like your call before touching commit history. The regen in commit 52c01a4 (`adcp: bump schema pin to 3.1.3`) does flip `CompletionRate` `float64` → `*float64` across ~12 delivery structs, and `adcp:` isn't a recognized conventional-commit type — so release-please won't route this as major. Options:

1. Amend the commit to `feat!: bump adcp schema pin to 3.1.3` with a `BREAKING CHANGE:` footer, force-push.
2. Add a new empty commit with the breaking marker in its message.

Option 1 is what release-please expects. Force-push required — let me know if you want me to do it.

PR #423, opened 2026-07-23
bench: split perf into identity-perf + context-perf; add reference results on main

ohalushchak-exadel:
## Summary

- **Rename** `bench/perf/` → `bench/identity-perf/` for symmetry with the new `bench/context-perf/` sibling. Mechanical: `go.mod` path, import in seeder+writer, Dockerfile COPY, compose Dockerfile refs, README/run.sh/`.gitignore` comments. No behavior change.
- **Add reference results** to `bench/identity-perf/README.md` from a fresh full-matrix sweep on clean `origin/main` (98 runs, ~50 min wall time).
- **Add `bench/context-perf/`** — a new harness mirroring identity-perf but adapted to the context-agent's request shape (ContextMatchRequest with `artifact_refs` / `context_signals` against `mediabuystore` + `pkgconfigstore` + `topicstore` + `signalstore`). Three scenarios (`packages-only`, `packages-topics`, `packages-signals`), same 7 × 7 CPU/mem × QPS sweep, same CSV shape so results roll up compatibly.
- **Reference results for context-perf** also included (147 runs, ~90 min wall time).

Both reference sweeps ran on the same host: AMD EPYC 9254 (24 c / 48 t, 4.15 GHz max), 384 GiB DDR5, Ubuntu 24.04.4, Docker 29.1.3 (compose v2.40.3), bare metal.

## context-perf headline numbers

Achieved QPS at target=32,000 (`_2g` / `_4g` / `_8g` variants of each CPU count matched to within noise; memory is not a constraint on any scenario):

| CPUs | packages-only | packages-topics | packages-signals |
|-----:|--------------:|----------------:|-----------------:|
| 1    | 1,929         | 1,406           | 1,094            |
| 2    | 4,090         | 2,803           | 2,244            |
| 4    | 6,132         | 4,266           | 3,763            |
| 8    | 6,652         | 4,972           | 5,103            |

Notable observations from the full run:
- Per-core efficiency peaks at **2 CPUs** on every scenario (same shape as identity-agent). Real diminishing returns past 4 CPU.
- **context-agent is ~3–4× more expensive per core than identity-agent** at comparable CPU allocations — expected: it evaluates 200 active packages per request including pkgconfig MGets + per-artifact topic / signal fan-out + suppression checks vs. identity-agent's single fcap MGet + audience HEXISTS.
- Memory: 37–100 MB RSS across the entire matrix regardless of config. **512 MB is enough, 1 GB is comfortable.**
- Errors: 0 transport errors, ≤ 11 non-2xx per 32k-target run across the whole 147-row matrix. All non-2xx are 1-CPU rows queueing past the 1 s `REQUEST_TIMEOUT` under saturation. Every ≥ 2-CPU row is clean.
- **`packages-signals` overtakes `packages-topics` at 8 CPU** (5,103 vs 4,972 qps) despite being slower at 1-4 CPU. Profiled: at 8 c, ~27% of ALL CPU is spent on `sync.Mutex.Lock/Unlock` + `golang-lru/v2 MoveToFront` on the topicstore LRU caches. The topics path does ~200 cache lookups per request (one per package), all fighting for the same two mutexes; the signals path does a single Redis MGet with no per-package cache lookups and scales freely with cores. Details in the PR discussion. Documented for whenever the topic path becomes a real bottleneck in production; not filing a ticket.

Full CSV + per-step JSON + `/metrics` snapshots archived on the bench host under `bench/context-perf/results/main-20260723T190414Z/`.

## identity-perf regression flagged

Cross-check between the fresh main run and the PR #413 baseline (same CPU model) shows a **consistent 15–25% regression on the fcap-only path and 30–40% on fcap-audience at CPU-bound configs**. `ok_2xx == total`, `non_2xx = errors = 0` in both datasets → not a correctness change reducing work; per-request cost went up.

Follow-up investigation posted as a PR comment: all five candidate commits between #413's baseline and current main are **ruled out** by code inspection (union-store wrapper is `Enabled`-gated and stays off in the bench env; audience fail-closed change is a nil-check on the success path; Gosched can't produce a CPU-cgroup-shaped regression; URL-scan is ~50 ns). Likely culprit is **base-image drift**: `cmd/identity-agent/Dockerfile:1` uses `golang:1.26-alpine` without a patch pin — a silent Go point-release between #413's build and my rebuild fits the uniform per-request-cost bump. Falsifier (checkout `c326342`, rebuild with current Go 1.26.5, rerun 4c/4g fcap-only @ 32k) is a follow-up. Hygiene fix (pin `golang:1.26.x-alpine` in all four Dockerfiles) is also a follow-up.

## Context on context-perf design choices

Notable divergences from identity-perf worth surfacing in review:
- `REQUEST_TIMEOUT` set to **1s** (production example.env uses 40ms) — deliberately generous so the deadline doesn't clip p99 during a saturation sweep.
- Signal `key_type` is **fixed to `url_hash`** in the seeder, not env-tunable — loadgen only emits `ArtifactRefType=url_hash`, so any override would silently seed keys the engine never reads.
- Signal-key count is **coupled to `TOTAL_ARTIFACTS`** — one signal per URL in the loadgen's pool. No separate `SIGNAL_VALUES` knob → impossible to decouple the two pools by accident.
- Every package AND every artifact carries a **shared always-on topic** (index 0) in the packages-topics scenario, so the topic gate warms the topicstore MGet + intersection path on every request instead of falling through to fail-closed on empty intersection.
- `perf.env` is the single source of truth for `PROPERTY_RIDS`; every service loads it via docker-compose `env_file:` so drift between seeder / loadgen / context-agent is impossible.

## Not (yet) in this PR — planned follow-ups

- **Signature-verify mode** for both harnesses — tracked as [AI-4641](https://linear.app/scope3-projects/issue/AI-4641/bench-add-signature-verify-mode-to-identitycontext-perf-harnesses).
- **Shared loadgen library** (`bench/loadgenlib/`) — pacer / percentile / envInt-Atoi helpers are still copy-pasted between the two harnesses. Extracting requires touching identity-perf's `go.mod` (adding a `replace` for the nested `bench/` module); deferred since identity-perf was actively being benchmarked during this PR's development.
- **Bisect the identity-perf regression** (build #413's `c326342` with the current Go 1.26.5 to confirm base-image drift as the cause). Pin exact Go patch versions in all four Dockerfiles either way.

## Test plan

- [x] Fresh `origin/main` checkout on latitude host, `bench/identity-perf/run.sh` clean run — 98 rows in summary.csv, `non_2xx = errors = 0` on every row.
- [x] Code review of `bench/context-perf/` (10 findings surfaced, all correctness ones fixed in this branch — see commit history).
- [x] `bench/context-perf/run.sh` clean run on latitude host — 147 rows in summary.csv, 0 transport errors, 79 non-2xx total (all in 1-CPU rows hitting the 1 s `REQUEST_TIMEOUT`).
- [x] Profiled the topics vs signals crossover at 8c — mutex contention on the topicstore LRU cache is the root cause, confirmed via CPU + goroutine profiles.
- [ ] Bisect the identity-agent regression between #413 and current main — separate PR.


ohalushchak-exadel:
## Regression investigation — code is exonerated, base-image drift is the likely cause

Traced the four suspects I named in the PR body:

| commit | verdict | why |
|---|---|---|
| `103cbdb` union-read shadow-mode fallback | **ruled out** | Wrapper is gated on `cfg.FallbackFCapValkey.Enabled` at `targeting/identityagent/setup.go:363` (audience twin at :391). Bench env sets no `*_FALLBACK_VALKEY_*` vars → `fcapStore` stays the primary `*redisstore.Store`. Zero request-path overhead. |
| `3f8d1a6` audience fail-closed on store error | **ruled out** | Adds one `if err != nil` per request in `EvaluateIdentityResolved` (identity_engine.go:66). Success path (`errors==0` in both datasets) is a single nil-check. |
| `d52e418`, `1a77335` | **ruled out** | Same `Enabled`-gated code paths as above, plus an error-path metric removal that's never hit. |
| `82c7995` control-char scan on `SellerAgentURL` | **weak (~50 ns/req)** | Adds a `for _, r := range value` in `ValidateIdentityRequest`. Nowhere near 15-25%. |
| `2a8a45c` loadgen `runtime.Gosched()` on drop | **ruled out** | Regression shape (worst at intermediate CPU, matches at 8c because 8c is loadgen-ticker-bound at 32k) is a SUT-cost story. A loadgen-side depression would be uniform across cgroup sizes. |

## Actual likely cause

`cmd/identity-agent/Dockerfile:1` — **`FROM golang:1.26-alpine`** — floating tag, no version pin.

- Current `golang:1.26-alpine` on the bench host resolves to **Go 1.26.5**.
- The #413 baseline was built ~24 h earlier from the same floating tag — a different Go patch could have been current then. A silent runtime/compiler change between patches is a much better fit for a uniform per-request-cost regression that scales with SUT CPU pressure.

## Falsifier (deferred until after the context-perf run completes)

Checkout `c326342` (PR #413's baseline commit), `docker rmi golang:1.26-alpine && docker pull golang:1.26-alpine` for reproducibility, rebuild the image, rerun **4c/4g fcap-only @ 32k**:

- If achieved_qps ≈ 26k → the code is exonerated, Go version drift is the cause.
- If achieved_qps ≈ 32k → some post-#413 commit is responsible after all and my analysis missed it.

## Hygiene follow-up either way

Pin exact Go patch versions in `cmd/{identity,context}-agent/Dockerfile` and `bench/{identity,context}-perf/Dockerfile` (e.g. `golang:1.26.5-alpine`) so future perf comparisons across time aren't confounded by a floating base image. Happy to include this in the current PR or as a separate one — small change either way.

PR #428, opened 2026-08-03
fix(router): honor provider priority for duplicate offers

bokelley:
## Summary

Implements the proposed AdCP 3.2 TMP merge rule from adcontextprotocol/adcp#5722:

- carry provider registration `priority` through discovery and Context Match fan-out
- select the lower numeric priority when providers return the same `package_id`
- retain first-response order only as the equal-priority tie-breaker
- preserve union semantics for Identity Match (adcontextprotocol/adcp#5721)
- log the selected provider and priority for duplicate-offer configuration errors

Companion spec ratification packet: adcontextprotocol/adcp#6170

## Verification

- `go test ./...`
- priority selection, equal-priority arrival tie-break, and registration conversion are covered directly

Closes #390's remaining priority-selection gap.

### DIFF
--- a/router/config.go
+++ b/router/config.go
@@ -63,14 +63,9 @@ type ProviderConfig struct {
 
 	Timeout time.Duration `json:"timeout"`
 
-	// Priority is documented in the schema (tmp/provider-registration.json) and
-	// the spec config sample. Accepted on the wire so spec-aligned configs
-	// parse, but has no effect on routing today: the upstream spec contradicts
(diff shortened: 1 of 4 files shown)


bokelley:
Triage status: valid and approved, but not merge-ready. The companion spec PR adcontextprotocol/adcp#6170 is merged; this implementation branch now conflicts with main. Please rebase onto current main, resolve the router/config drift, and rerun the full root-module tests. It can return to the merge queue after green CI.

